PR #4, opened 2015-10-15
Small fix and a failing test for the core meta schema draft 04

pertyjons:
Fixed a small error that the $ref has to start with a '#' but it only checks if it's one '#' and in the  core meta schema the path[0] will be '##' as the schema ends with a #. Maybe that should be solved in another way?

Adding a test failing test for validating the http://json-schema.org/draft-04/schema# schema with a StackOverFlow Exception due to a recursive loop when downloading the schema. Just to show the error.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/loader/internal/JSONPointer.java
+++ b/core/src/main/java/org/everit/json/schema/loader/internal/JSONPointer.java
@@ -153,7 +153,7 @@ public class JSONPointer {
       return new QueryResult(document, document);
     }
     String[] path = fragment.split("/");
-    if (!"#".equals(path[0])) {
+    if ( path[0] == null || !path[0].startsWith("#")) {
       throw new IllegalArgumentException("JSON pointers must start with a '#'");
     }
     Object current = document;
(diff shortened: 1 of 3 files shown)


erosb:
@pertyjons I checked out your fork, running `mvn clean install -U` resulted in StackOverflowError.

pertyjons:
Yes I know. I made the test to show  the error, I actually didn't fix the SOE. I made it as a heads up as I didn't had the time to dig into it at that time. 

I'm no expert in the JSON Schema inner workings but as the schema uses itself as a schema I suppose it get a recursive endless loop downloading itself and then SOE.

A simple check/cache that the schema is downloaded should probably be enough (without having looking into it to much).

erosb:
> A simple check/cache that the schema is downloaded should probably be enough

@pertyjons thats how it is meant to work (more or less): if a JSON pointer is found then a `ReferenceSchema.Builder` instance is first created for it, stored in the `SchemaLoader#pointerSchemas` map, then the actual schema loading happens. But it works only for relative pointers (in-document pointers - you may have a look at https://github.com/everit-org/json-schema/blob/master/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java#L536 if you feel curious). The JSON meta-schema has an `id` property therefore all pointers will be absolute pointers, thats why the loader ends up in an infinite recursion.

I'll dig into it

erosb:
fixed it here: https://github.com/everit-org/json-schema/pull/5

pertyjons:
@erosb Fast update :-D I really liked the clean Java 8 approach of the project so thank YOU.

PR #19, opened 2016-01-12
Modification for being Java 6 compliant

rdruilhe:
This branch has been refactored to be Java 6 compliant. It is based on the commit 3815c80, so it is not up-to-date with the master.

The changes mainly use the old style of coding of Java 6. The pattern Consumer/function/supplier is also created because it does not exist in Java 6.


erosb:
Uhh, sad news you had to put this level of effort into backporting. Well, I won't merge it, because it would need way too much effort to maintain a java6 branch of the library. I hope the java6/7 compatibility problem to fade away as time is passing.

Anyway, kudos for working on it :)

rdruilhe:
No problem, it was just for information because someone asked for it.

Issue #21, opened 2016-01-24
minProperties constraint ignored when using definitions/references in JSON schema


Antares42:
I have a JSON schema that defines a type ``language_object`` with some properties and then uses that on one of the fields, along with a ``minProperties`` constraint.

However, when validating against this schema, submitting an empty object in said field does not raise a ``ValidationException``. It appears that the ``minProperties`` constraint is ignored when it is placed alongside ``$ref``

I have prepared a schema that defines my field once as ``name`` using references and once as ``nameEx`` with explicit properties. As far as I can tell these should be equivalent:

	{
	    "type":"object",
	    "$schema":"http://json-schema.org/draft-04/schema#",
	    "definitions":{
	        "language_object":{
	            "type":"object",
	            "additionalProperties":false,
	            "patternProperties":{
	                "^[a-z]{2}$":{
	                    "type":"string"
	                }
	            }
	        }
	    },
	    "properties":{
	        "name":{
	        	"$ref":"#/definitions/language_object",
	            "minProperties":1
	        },
	        "nameEx":{
	        	"type":"object",
	            "patternProperties":{
	                "^[a-z]{2}$":{
	                    "type":"string"
	                }
	            },
	        	"additionalProperties":false,
	            "minProperties":1
	        },
	        "date":{
	            "type":"string",
	            "format":"date-time"
	        }
	    }
	}

When I validate the following JSON

    {
       "name":{},
        "nameEx":{}
    }

I only get ``#/nameEx: minimum size: [1], found: [0]``, not both. Other situations, like this disallowed key/property work fine on both:

    {
       "name":{"nbb":"bla"},
       "nameEx":{"nbb":"bla"}
    }

Returns

    #: 2 schema violations found
    #/nameEx: extraneous key [nbb] is not permitted
    #/name: extraneous key [nbb] is not permitted

Thank you for your patience!

Antares42:
I can fork and create a pull request if you like.

erosb:
@Antares42 I fixed it in 966e9bb4a6a3f590c162cfc0bd8c523bfc5e9aee .

I'd also like to add support for the "format" and "default" keywords. So now I'm going to look after how much work do these need. If it seems to be a lot then we will release 1.1.2 with this single bugfix. If I can get "format" and "default" done this week then we will release these together in 1.2.0.

Antares42:
Cool! Your commitment to this project is impressive.

erosb:
[...]

PR #23, opened 2016-01-24
Issue 21: minProperties:1 but no ValidationException when sending {}

Antares42:
[Issue Link](https://github.com/everit-org/json-schema/issues/21)


erosb:
Thanks for the testcase, merged.

PR #39, opened 2016-06-21
Allow users to specify expected values for validation failures

twbutler:
As discussed in Issue 36, I implemented the ability for a user to add an issue testcase and specify the exact validation failures that they are expecting to be generated by the library.  Along with these changes, I am submitting three new files under issue36/ that demonstrate the new functionality.  Note that the subject JSON file contains 4 validation failures. The library correctly finds all 4, but the top level summary message incorrectly states "3 validation failures".  This is a potential bug, to be determined by the maintainer based on the design of the library... see what you think. 

See the TODO comments also.  These are utility methods I added to IssueTest (processValidationFailures(), loadJsonFile()) that could be moved into the core project, so that users could use them as convenience methods.  If this were done, processValidationFailures() would need to be modified to return a List (of the validation failure strings).  Again, you can see what you think...

Regards,
Trevor

### DIFF
--- a/tests/src/test/java/org/everit/json/schema/IssueTest.java
+++ b/tests/src/test/java/org/everit/json/schema/IssueTest.java
@@ -29,6 +29,7 @@ import java.util.Optional;
 
 import org.eclipse.jetty.server.Server;
 import org.everit.json.schema.loader.SchemaLoader;
+import org.json.JSONArray;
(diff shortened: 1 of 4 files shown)


erosb:
Thank you for implementing it!
It may need some slight modifications, but I will handle those, and I will also keep your suggestions in mind.

PR #43, opened 2016-07-27
Error Message

fernandostoiano:
What do you think about that method? Because if you need the super.message() is important for get that message more easy.

For example in my project I need the super.message and I don't have access. If I try implement one solution in my project is impossible because I don't have a pattern for take a substring.


erosb:
Thank you, merged.

Issue #44, opened 2016-08-05
Validate if field exists in schema


v1ctor:
Hi!
I would like to add new feature to validate if field exists in schema.
```java
Schema schema = SchemaLoader.load(schemaAsJsonObject);
Assert.assertTrue(schema.hasField('rectangle.a'))
```
In first glance I think that I have to validate only propertySchemas recursievly and patternProperties.
Could you please point if there are some corner cases of this solution?

v1ctor:
Hi @erosb You could close this task :)
Also could you please tell me when you will release new version of the library?

erosb:
Hi! After merging your PR I recognized that your ObjectSchema#definesProperty() implementation handles escaping & recursion only for properties, but not for patternProperies and schemaDependencies. I started working on it. With some luck I will finish it tonight.

erosb:
Okey, all looks fine, the new release is [available on GitHub](https://github.com/everit-org/json-schema/releases/tag/1.4.0).
@balazs-zsoldos please deploy version 1.4.0 to maven central , thanks.

PR #45, opened 2016-08-05
44: Add possibility to check if field exists in schema

v1ctor:
#44

### DIFF
--- a/core/src/main/java/org/everit/json/schema/ObjectSchema.java
+++ b/core/src/main/java/org/everit/json/schema/ObjectSchema.java
@@ -395,7 +395,7 @@ public class ObjectSchema extends Schema {
   }
 
   @Override
-  public boolean hasField(String field) {
+  public boolean definesProperty(String field) {
     if (field.isEmpty()) {
       return false;
     }
@@ -410,7 +410,7 @@ public class ObjectSchema extends Schema {
     if (propertySchemas.containsKey(current)) {
       if (hasSuffix) {
         String suffix = fields.get(1);
-        return propertySchemas.get(current).hasField(suffix);
+        return propertySchemas.get(current).definesProperty(suffix);
       } else {
         return true;
       }
--- a/core/src/main/java/org/everit/json/schema/ObjectSchema.java
+++ b/core/src/main/java/org/everit/json/schema/ObjectSchema.java
@@ -396,14 +396,15 @@ public class ObjectSchema extends Schema {
 
   @Override
   public boolean definesProperty(String field) {
+    field = field.replaceFirst("^#", "").replaceFirst("^/", "");
     return !field.isEmpty() && (definesPatternProperty(field)
             || definesSchemaDependencyProperty(field)
             || definesSchemaProperty(field));
   }
 
   private boolean definesSchemaProperty(String field) {
-    List<String> fields = Lists.newArrayList(Splitter.on(".").limit(2).split(field));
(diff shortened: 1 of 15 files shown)


erosb:
Hello, basically don't mind the idea, but I have a number of concerns regarding the implementation:

* an other name must be chosen, I find the current one misleading. Maybe "definesProperty".
* the dot notation of properties is something I'd like to avoid. It would be better to use JSON pointers, since that is a standard notation for the same purpose. It also handles escaping.
* ObjectSchema#schemaDependencies must be handled, in addition to properties and patternProperties
* the current implementation of NotSchema#hasField() is confusing. I will think about it
* Adding a new abstract method to Schema is not a good idea, since it breaks backward-compatibility for people who created their own subclasses (it may be a rare usecase, but still). A default implementation returning "false" would be better,  since you could remove most of the overrides.

Sent from my phone, sorry about the typos.
Cheers

erosb:
@v1ctor are you going to continue working on this?

v1ctor:
@erosb yes, sure. I will verify schemaDependencies and change format to pointer notation soon.

erosb:
Awesome. Please go on with `schemaDependencies`, and I'd like to do something about `NotSchema#definesProperty()` too. Do you have any thoughts? Maybe it should always return `false`.

Please wait a bit with the JSON Pointer notation. The problem there is that we already have some sort of duplication in the library regarding JSON pointers, since
* we already have a `org.everit.json.schema.loader.internal.JSONPointer` class, which is not too useful for this task
* there is also some code for pointer notation handling in `ValidationException#escapeFragment()`
* and in the upcoming version of the `org.json:json` library there is also a [JSON Pointer implementation](https://github.com/stleary/JSON-java/issues/218)

So I think the best would be to wait for the next release of the org.json lib ([coming soon](https://github.com/stleary/JSON-java/issues/250)) and use that class for this purpose. I will also try to remove the my half-baked implementations from this schema library.

v1ctor:
I've written code to validate schemaDependencies and decided to keep default realisation in NotSchema.

core/src/main/java/org/everit/json/schema/ObjectSchema.java
```diff
@@ -418,4 +416,26 @@ public boolean definesProperty(String field) {
     return false;
   }
 
+  private boolean definesPatternProperty(String field) {
```
erosb: You can rewrite this method to use the java8 stream API. It isn't a necessary change, but it could be good for practicing  it (if you are not familiar yet with streams), due to the simplicity of the task. The same goes for `definesSchemaDependencyProperty()`

erosb:
Great job. In the meanwhile the new version of `org.json:json` is has been published to maven central, but its `JSONPointer` class doesn't seem to be useful for this scenario. Please go on with changing the dot notation to JSON Pointer notation. The method should accept a JSON pointer in its [string representation](https://tools.ietf.org/html/rfc6901#section-5). Please follow the RFC, and write some tests for verifying the un-escaping of the `~0` and `~1` tokens. Basically you have to do the opposite of what `ValidationExcepion#escapeFragment()` does. I'm looking forward to merge this PR quite soon, and release 1.4.0 this week.

v1ctor:
@erosb Thanks! I've changed realisation to use JSONPointer notation.

erosb:
All looks fine, I will add some javadoc, then it is ready to be released. Thank you for your contribution.

PR #50, opened 2016-08-29
add equals and hashcode for schema objects

### LINKED ISSUE #49: Schema types do not implement equals
The different schema types (StringSchema, ArraySchema, etc) do not implement equals() and so cannot be compared to each other.

tylersouthwick:
This makes the derived schema classes final. This could break backward compatibility. If needed, that can be relaxed.

fixes #49

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -100,5 +100,11 @@
       <version>1.10.19</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>nl.jqno.equalsverifier</groupId>
+      <artifactId>equalsverifier</artifactId>
+      <version>2.1.5</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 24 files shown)


erosb:
Please remove the added `final` modifiers from the classes. Maybe it wasn't a good decision originally to keep them extendable, but at this point making them final would be a slightly backward-compatibility breaking change.

tylersouthwick:
I removed final from the classes. There are some different ways of making the equals/hashCode valid with non final classes- ideally we could make equals/hashCode final, but that could also break backwards compatability. There are some improvements on the tests that would be nice to have more solid equals/hashCode tests - for now, I'm having it suppress warnings for subclass equality as I don't know if that is a use case worth investing in. What are your thoughts?

erosb:
Okey, all looks good. Java, and OOP in general has some defect regarding equality. Resolving it is out of the scope of this PR :)
Thank you for your efforts.

PR #54, opened 2016-11-14
Add ValidationException#getAllMessages() for collecting nested error messages

srfarley:
`ValidationException#getAllMessages()` walks through all nested violations and return the list of error messages. This is particularly useful for use by a REST API when validating the schema of an incoming JSON payload. The flat list of messages provides a convenient format for clients to log, and even parse, the errors.

I also added a unit test to verify existing and new behavior for multiple nested schema violations.


erosb:
Merged. Thank you for your work @srfarley !

srfarley:
@erosb You're quite welcome! Do you have a date planned for the next release?

erosb:
Definitely this year, but I don't know any more accurate release date.

PR #58, opened 2016-11-25
Fix unicode length comparisons

halvorgb:
The test suite had the wrong expected values for max/minLength for unicode strings, they've been updated. See: [draft4/minLength.json#29](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/develop/tests/draft4/minLength.json#L29) and [draft4/maxLength#29](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/develop/tests/draft4/maxLength.json#L29)

Use the number of unicode codepoints to compare lengths.  see: [stackoverflow](http://stackoverflow.com/questions/6828076/how-to-correctly-compute-the-length-of-a-string-in-java)

Why was the test suite logic flipped? Old data? Perhaps looking into importing the test suite as git submodule would be a good idea?

### DIFF
--- a/core/src/main/java/org/everit/json/schema/StringSchema.java
+++ b/core/src/main/java/org/everit/json/schema/StringSchema.java
@@ -131,7 +131,7 @@ public class StringSchema extends Schema {
     }
 
     private List<ValidationException> testLength(final String subject) {
-        int actualLength = subject.length();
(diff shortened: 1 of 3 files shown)


erosb:
Thank you, merged.

Probably the repository contained these misleading testcases because at the very beginning of the project I temporarily changed them, then I forgot to restore.

PR #59, opened 2016-11-25
Implement variable length seconds fraction to conform with RFC-3339

deepakhalale:


core/src/test/java/org/everit/json/schema/internal/DefaultFormatValidatorTest.java
```diff
@@ -44,18 +44,64 @@ private void assertSuccess(final String subject, final FormatValidator format) {
...
+    }
+
+    @Test
+    public void dateTimeWithNineDigitsInSecFracSuccess() {
+        assertSuccess("2015-02-28T11:00:00.123456789Z", new DateTimeFormatValidator());
     }
 
     @Test
-    public void dateTimeSecFracSuccess() {
-        assertSuccess("2015-02-28T11:00:00.111Z", new DateTimeFormatValidator());
```
deepakhalale: Hey @erosb, don't think I have removed the test case. I have renamed it to dateTimeWithThreeDigitsInSecFracSuccess and updated the seconds fraction from "111" to "123" to aligh with rest of the test cases that I added.
If you think the test case was there for a reason, then I will be glad to put it back.

erosb: Ok, found it.
One final note: I've just found in the [RFC3339 document](https://www.ietf.org/rfc/rfc3339.txt) that their ABNF defines fractions as:

    time-fraction     = ("," / ".") 1*DIGIT

I had a quick glance at the javadoc and I couldn't find any easy way to support comma too, not only dot as a separator character. Do you think there is any simple solution to fix it?

deepakhalale: The [Json Schema Validation spec](http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1) seem to suggest that the date-time format need to specifically conform to [Section 5.6 of RFC-3339](https://tools.ietf.org/html/rfc3339#section-5.6). 
I think Section 5.6 of RFC-3339 is a specific cut down version of the ISO 8601 grammar and the [ABNF date time](https://tools.ietf.org/html/rfc3339#appendix-A) that your referring to in the above comment, is an ISO 8601 standard.
Hope it helps.

erosb:
All looks fine, merging. Thank you for your contribution @deepakhalale !

PR #67, opened 2016-12-28
some code clean up

agebhar1:
* removed unused import(s)
* removed unused field(s)


erosb:
Thank you @agebhar1 !
Probably I should configure my IDE to do it automatically.

agebhar1:
Your are welcome!

... it seems, you're using Intelli J - the organization of the import block in Eclipse is not the same by default settings :unamused: 

For Intelli J have a look on the [Save Actions](https://plugins.jetbrains.com/idea/plugin/7642-save-actions) plugin.

PR #111, opened 2017-10-10
Updating Maven information based on jitpack.io

smarkwell:
Based on this https://jitpack.io/#everit-org/json-schema/1.6.0 I've updated the Maven link to reflect what appears to be the correct

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ Add the JitPack repository and the dependency to your `pom.xml` as follows:
 
 ```xml
 <dependency>
-    <groupId>com.github.everit-org.json-schema</groupId>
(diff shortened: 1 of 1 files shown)


erosb:
Hi! I might overlook something, but the current groupId/artifactId seems to be correct for me, according to the link you referenced:

screenshot from 2017-10-10 07-08-19

smarkwell:
Interesting, when I use the following URL I get my result
'https://github.com/everit-org/json-schema'

image

smarkwell:
Another screen shot

image

erosb:
That is strange.

Anyway, when I try to build a skeleton project with `com.github.everit-org:json-schema` then it fails to build.

Issue #113, opened 2017-10-18
format over object properties


cfcfs:
I am using this library to validate a large configuration document. For some parts of it using `propertyNames` is not enough to fully validate the property names, so I was looking into `format` for that. 
The library doesn't  support it, but adding that support doesn't seem to be hard. Would you accept a pull request on this?

Thanks!

erosb:
Closing this, I added the explanation to #114 .

PR #114, opened 2017-10-18
add format validation to object schema

cfcfs:
Issue #113


erosb:
Hello @cfcfs ,

thank you for putting effort into this PR. At the first glance I don't clearly understand the purpose of this PR. It is a short-hand notation for 

```
{
    "propertyNames": { "format" : "..." }
}
```
Or am I overlooking something here?

cfcfs:
Hello @erosb 

I'm new to json-schema and totally missed `propertyNames`. So yes, using `format` on a object would be equivalent to defining it on its `propertyNames`. 
Despite that, I would expect `format` to work this way on object schemas and it saves a lot of typing for large documents. 
I can live with the current form though, If you think this doesn't make sense.

Thanks.

erosb:
Hello @cfcfs ,

thank you for your response. Unfortunately it doesn't seem to be a good reason for me to merge this PR. The main problem is that *it is not a standard feature defined by the json-schema specification*, and I don't find it a good idea to support non-standardized features (because users will start depending on it, then they are surprised if the feature doesn't work with other validator implementations).

It is a quite unlucky situation. I really appreciate your PR, it was seemingly written with care and awareness. Please wait for some discussion next time before you start coding, so we don't waste your time.

Issue #116, opened 2017-10-31
StringsSchema has not implemented 'not'.


kazie:
Whoever it might concern.

While using the latest Everit Json-schema validator, I happen to notice that it didn't let me run though on strings where I had implemented "not enum" blocks, or other "not" blocks for that matter.

Examples Schema:
```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string",
      "not": {
        "type": "string",
        "pattern": "^([0-9a-fA-F][0-9a-fA-F])+$"
      }
    }
  },
  "required": [
    "name"
  ]
}
```
Didn't invalidate 
```json
{
 "name" : "DEAD"
}



Or in the enum case.
```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string",
      "not": {
        "enum": [
          "A",
          "B",
          "C"
        ]
      }
    }
  },
  "required": [
    "name"
  ]
}
```

Didn't invalidate:
```json
{
   "name": "A"
}
```

As one can test one self, should be valid in other validators, such as the online validator on:
https://jsonschemalint.com/#/version/draft-06/markup/json

----

I will soon propose suggested fix, which also validates the changes by:
1) Not breaking old tests.
2) Validating new tests cases as described above.

----

Please if any comments, you can either "fix-it-yourself" or tell me if I need some changes in the code for a proposed pull request that will come with this ticket.

Best Regards,

erosb:
Hello @kazie ,

first, if you rewrite the schema to this form, then it will work:

```
{
  "type": "object",
  "properties": {
    "name": {
      "allOf" : [
         {
            "type": "string"
         },
         {
            "not": {"enum": ["A","B","C"]}
         },

      ]
    }
  },
  "required": [
    "name"
  ]
}
```

Second, please don't fix it as-is.

The library makes the false assumption that the keywords are / can be "grouped" in a specific type-related way. When the schema gets loaded, the loader makes some investigation about the possible type of the schema, then it loads the schema by taking only the schema-specific keywords into account.

In your case, when the loader finds `"type":"string"` , it checks further only the string-related keywords (like `"pattern"`, `"format"`, `"minLength"`, etc..). It does not look for `"not"` because it isn't considered as a string-related keyword, and it is implemented in `NotSchema` (so `"not"` works only if the schema has only a single `"not"` key).

This has been the most common reason why issues are raised for this library, so probably it would make sense to attempt to fix it.

The best thing I can imagine is
[...]

PR #117, opened 2017-10-31
Fixing #116, 'not' in strings.

kazie:
This fix will clear the issue that I got when using "not" blocks for validation of string.
    
Example would be to add forbidden values (not enum) or anti-patterns (not pattern).

Implementation should be quite straight forwards, and be implementing "not" as a complete schema of its own - using the existing "NotSchema" class to validate the schema.


erosb:
Thank you for your efforts, but I can't merge it. It is just a symptom of a deeper and more general problem. I added my explanation here: https://github.com/everit-org/json-schema/issues/116#issuecomment-340783568

Issue #91, opened 2017-07-12
default value support?


mrwilby:
We're exploring the use of this framework for programmatic schema generation, i.e. generating schemas using the builder/fluent API exposed as part of this library.

Does the library support default property values? I couldn't seem to find this but I may be overlooking it.

Thanks!

erosb:
Hello @mrwilby ,

currently the library doesn't support the `"default"` keyword. Although it is the part of the specification, a lot of implementations don't support it, and I personally didn't find it a good idea to mix validation of the input with modification (populating with defaults).

Anyway, we can keep this issue open, as a feature request.

slinkydeveloper:
Hi! I need it too, maybe if you add this feature and enable it with a flag?

erosb:
Hello @mrwilby , default values are now supported in version 1.7.0.

ganpa:
@erosb the schema allows default values to be invalid. Any reason for this ? Or do you plan to support validations for default values in the future ?

erosb:
Hello @ganpa , the schema allows default values to be invalid because this is what the JSON Scema specification says.

Citing [section 10.2](http://json-schema.org/latest/json-schema-validation.html#rfc.section.10.2):

> This keyword can be used to supply a default JSON value associated with a particular schema. It is RECOMMENDED that a default value be valid against the associated schema.

ganpa:
Yes. It says it is recommended and not must. But it would be great if the library validates default values too.

PR #123, opened 2017-12-04
Added default support

slinkydeveloper:
Solves https://github.com/everit-org/json-schema/issues/91 and can be used only with a flag :smile:

### DIFF
--- a/core/src/main/java/org/everit/json/schema/loader/ArraySchemaLoader.java
+++ b/core/src/main/java/org/everit/json/schema/loader/ArraySchemaLoader.java
@@ -40,6 +40,8 @@ class ArraySchemaLoader {
         ls.schemaJson().maybe("minItems").map(JsonValue::requireInteger).ifPresent(builder::minItems);
         ls.schemaJson().maybe("maxItems").map(JsonValue::requireInteger).ifPresent(builder::maxItems);
         ls.schemaJson().maybe("uniqueItems").map(JsonValue::requireBoolean).ifPresent(builder::uniqueItems);
+        if (config.useDefaults)
+            ls.schemaJson().maybe("default").map(JsonValue::deepToOrgJson).ifPresent(builder::defaultValue);
         ls.schemaJson().maybe("additionalItems").ifPresent(maybe -> {
             maybe.canBe(Boolean.class, builder::additionalItems)
                     .or(JsonObject.class, obj -> builder.schemaOfAdditionalItems(defaultLoader.loadChild(obj).build()))
--- a/core/src/main/java/org/everit/json/schema/loader/ArraySchemaLoader.java
+++ b/core/src/main/java/org/everit/json/schema/loader/ArraySchemaLoader.java
@@ -40,8 +40,6 @@ class ArraySchemaLoader {
         ls.schemaJson().maybe("minItems").map(JsonValue::requireInteger).ifPresent(builder::minItems);
         ls.schemaJson().maybe("maxItems").map(JsonValue::requireInteger).ifPresent(builder::maxItems);
         ls.schemaJson().maybe("uniqueItems").map(JsonValue::requireBoolean).ifPresent(builder::uniqueItems);
-        if (config.useDefaults)
-            ls.schemaJson().maybe("default").map(JsonValue::deepToOrgJson).ifPresent(builder::defaultValue);
         ls.schemaJson().maybe("additionalItems").ifPresent(maybe -> {
             maybe.canBe(Boolean.class, builder::additionalItems)
                     .or(JsonObject.class, obj -> builder.schemaOfAdditionalItems(defaultLoader.loadChild(obj).build()))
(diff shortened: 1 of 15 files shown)


erosb:
Hello @slinkydeveloper ,

thank you for putting effort into this. At the first glance my most serious concern about this PR is that it handles default values only if it is a member of an object, and there are some cases when the root value is not an object. For example if your root schema is
```
{
  "type": "string",
  "default": "empty"
}
```
and the input json is `null`, then probably the library should signal back `"empty"` as a default value somehow to the caller side. But that raises yet another question - how to provide the default value back to the caller? The signature of the `void validate(Object)` method doesn't provide any way, but changing it to return some data holder would be a backward-compatibility breaking change.

On the other hand, we can also ignore this impedance, and say that with this patch we cover the vast majority of usecases of the "default" keyword, so it is good-enough.

I will think about it.

I'm going to leave some comments in code review too, soon.

core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -178,6 +180,11 @@ SchemaLoaderBuilder pointerToCurrentObj(List<String> pointerToCurrentObj) {
             return this;
         }
 
+        SchemaLoaderBuilder useDefaults(boolean useDefaults) {
```
erosb: Would you please add some javadoc? Also, this seems to be the toggle point of the feature, so it would be necessary to make the method `public`, otherwise users won't be able to use this feature. I know there are some builder methods here which are package-private, but unlike this one, those are only for internal use by the library.

core/src/main/java/org/everit/json/schema/loader/JsonValue.java
```diff
@@ -234,4 +234,17 @@ public Integer requireInteger() {
         }
         throw ls.createSchemaException(typeOfValue(), Integer.class);
     }
+
+    public static Object deepToOrgJson(JsonValue v) {
```
erosb: Please make this package-private. I can see how is this necessary for now, but it is a bit redundant with `JsonValue#value()` and `JsonValue#unwrap()` (and their overrides in `JsonObject` and `JsonArray`) so this should be cleaned up in the future (and if you make this `public` then it can't be deleted in any 1.x versions).

core/src/main/java/org/everit/json/schema/loader/ArraySchemaLoader.java
```diff
@@ -40,6 +40,8 @@ public ArraySchemaLoader(LoadingState ls, SchemaLoader defaultLoader) {
         ls.schemaJson().maybe("minItems").map(JsonValue::requireInteger).ifPresent(builder::minItems);
         ls.schemaJson().maybe("maxItems").map(JsonValue::requireInteger).ifPresent(builder::maxItems);
         ls.schemaJson().maybe("uniqueItems").map(JsonValue::requireBoolean).ifPresent(builder::uniqueItems);
+        if (config.useDefaults)
+            ls.schemaJson().maybe("default").map(JsonValue::deepToOrgJson).ifPresent(builder::defaultValue);
```
erosb: See my notes in `#buildBooleanSchema()`

core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -404,8 +423,13 @@ public SchemaLoader(SchemaLoaderBuilder builder) {
         }
     }
 
+    private BooleanSchema.Builder buildBooleanSchema() {
```
erosb: I think now you can remove this method (just return `BooleanSchema.builder()` in line 414)

core/src/test/resources/org/everit/jsonvalidator/testschemas.json
```diff
@@ -543,5 +543,44 @@
         }
       }
     }
+  },
+  "defaultsTest": {
```
erosb: Would you please also add a test with `"default": null` ? The tricky thing there is that the org.json library represents null values with `JSONObject#NULL` instead of a null reference, so maybe you will have to improve the `#deepToOrgJson()` method to cover it.

core/src/main/java/org/everit/json/schema/loader/StringSchemaLoader.java
```diff
@@ -18,25 +18,30 @@
...
         this.formatValidators = unmodifiableMap(requireNonNull(formatValidators, "formatValidators cannot be null"));
+        this.useDefault = useDefault;
     }
 
     public StringSchema.Builder load() {
         StringSchema.Builder builder = StringSchema.builder();
         ls.schemaJson().maybe("minLength").map(JsonValue::requireInteger).ifPresent(builder::minLength);
         ls.schemaJson().maybe("maxLength").map(JsonValue::requireInteger).ifPresent(builder::maxLength);
+        if (useDefault)
+            ls.schemaJson().maybe("default").map(JsonValue::requireString).ifPresent(builder::defaultValue);
```
erosb: Now this can be removed (you unified the loading in `SchemaLoader#load()`)

PR #124, opened 2017-12-04
Added nullable support

slinkydeveloper:


core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -365,6 +372,8 @@ public SchemaLoader(SchemaLoaderBuilder builder) {
         ls.schemaJson().maybe(config.specVersion.idKeyword()).map(JsonValue::requireString).ifPresent(builder::id);
         ls.schemaJson().maybe("title").map(JsonValue::requireString).ifPresent(builder::title);
         ls.schemaJson().maybe("description").map(JsonValue::requireString).ifPresent(builder::description);
+        if (config.specVersion == OAS_3_SCHEMA)
+            ls.schemaJson().maybe("nullable").map(JsonValue::requireBoolean).ifPresent(builder::nullable);
```
erosb: Please add the braces around the statement.

core/src/test/java/org/everit/json/schema/loader/SchemaLoaderTest.java
```diff
@@ -484,4 +467,30 @@ public void relativeIdInReferencedSchemaRoot() {
...
+        JSONObject rawSchema = ALL_SCHEMAS.getJSONObject("nullableTest");
+        ObjectSchema schema = (ObjectSchema) SchemaLoader.builder().schemaJson(rawSchema).OAS3SchemaSupport().build().load().build();
+
+        try {
+            schema.validate(new JSONObject("{\"hello\": \"world\", \"world\": null}"));
+            assertTrue("Exception not thrown!", false);
+        } catch (ValidationException e) {
+            assertEquals(true, true);
+        }
+    }
```
erosb: The convention is that SchemaLoader unittests only verify that the proper Schema object is constructed from the raw json, so in this case, in the  `SchemaLoaderTest` you should just make assertions on the `getNullable()` flag of the created Schema instance.

Then to check if the `nullable` flag is properly handled by the Schema subclasses, you can add a new testcase for each `Schema` subclass to check if the nullable flag is taken into account properly (that means one check for each `checkNullity()` call). For these tests, please use the `TestSupport` class which helps you to expect conventional `ValidationException` instances (which is a bit more detailed than exceptions usually are).

core/src/main/java/org/everit/json/schema/Schema.java
```diff
@@ -221,4 +238,12 @@ protected ValidationException failure(Class<?> expectedType, Object actualValue)
     protected boolean canEqual(final Object other) {
         return (other instanceof Schema);
     }
+
+    protected boolean checkNullity(final Object subject) {
```
erosb: This implementation looks odd to me. This should be a void method and it should do something like `throw failure("value is null", "nullable");` instead of returning false.

I played around with it locally a bit, and it seems the `SchemaLoaderTest#nullableTestError()` passes only by accident (the exception message is "expected type: String, found: Null"), and wouldn't work for schemas which don't declare their `"type"` explicitly (which is valid in json schema).

Please change this method to throw an exception instead of returning `false`, and also write some unittests for each Schema subclasses using it (as mentioned below).

slinkydeveloper: There's no other way to do that. I've tried to do as you wish, but in fact I need a way to "skip" the remaining validation process after I've found that something is nullable and is null. If the validation goes through, It gives an error when it checks the type (it founds type NULL, expecting for example a String)

Issue #126, opened 2017-12-16
Enum: Validating allowed null value fails


heneke:
Consider the following enum schema:

`"enum": [null, "foo", "bar"]`

A valid null value in the JSON instance will be considered invalid, because the null is not added to the set of possible values.

heneke:
Sorry, it is added, but validation fails nevertheless.

heneke:
See https://github.com/everit-org/json-schema/pull/127

erosb:
Fixed in 60cbcca1887949f5935b6989d5a579746330cc2f .

PR #127, opened 2017-12-17
Comparision of null values to allow null values in enum schemas

heneke:


erosb:
Hi! I've just pushed an other fix of #126 into `master` : 60cbcca1887949f5935b6989d5a579746330cc2f .

I find it a bit more appropriate than yours (bit more tests and a bit better place for the fix), although this PR would also be sufficient.

I'm closing this PR. Do you need an urgent release with this fix? If you don't need any specific release number then you can just use the fixed version by referring to the commit hash of the fix:

```
    <dependencies>
        <dependency>
            <version>60cbcca1887949f5935b6989d5a579746330cc2f</version>
        </dependency>
```

heneke:
No offense taken – you know the codebase way better than me.

PR #131, opened 2018-01-07
Time format validator

zgyorffi:


core/src/main/java/org/everit/json/schema/internal/TemporalFormatValidator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.time.temporal.ChronoField;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Base class for date and time format validators
+ */
+public class TemporalFormatValidator implements FormatValidator {
+    private DateTimeFormatter formatter;
+    private List<String> formats_accepted;
```
erosb: Thanks for the update @zgyorffi .
Please declare the instance-level fields below the static members. Also, please use camelCase naming instead of under_score (so please rename `formats_accepted` to `formatsAccepted` both here and in the constructor).
Otherwise it looks good now.

core/src/main/java/org/everit/json/schema/internal/TemporalFormatValidator.java
```diff
@@ -0,0 +1,39 @@
...
+
+    final static DateTimeFormatter SECONDS_FRACTION_FORMATTER = new DateTimeFormatterBuilder()
+            .appendFraction(ChronoField.NANO_OF_SECOND, 1, 9, true)
+            .toFormatter();
+    static final String ZONE_OFFSET_PATTERN = "XXX";
+
+    TemporalFormatValidator(DateTimeFormatter formatter, List<String> formats_accepted) {
+        this.formatter = formatter;
+        this.formats_accepted = formats_accepted;
```
erosb: Minor nitpicking here: would you please add a null check for `formatter` (so ` this.formatter = requireNonNull("formatter", "formatter cannot be null";`) and defensively copy the `formatsAccepted` list? (so `this.formatsAccepted = new ArrayList<>(formatsAccepted);`)
Thanks again.

PR #133, opened 2018-01-11
ARUHA-1275 Fix URI validation

### LINKED ISSUE #132: Is the URI validator for draft 4 too strict?
It seems to me that the URI validator for draft 4 is too strict since version 1.6.0

In version 1.5.1, this string would be considered a valid URI: `abc`. Since version 1.6.0, this seems not to be the case anymore, though `abc` is accepted by, eg, `java.net.URI`.

It is true that the json-schema spec refers to RFC3986, while the `java.net.URI` javadoc refers to RFC2396 and RFC2732.
RFC3986, which obsoletes the other two, clarifies the following:

      NOTE: Previous specifications used the terms "partial URI" and
      "relative URI" to denote a relative reference to a URI.  As some
      readers misunderstood those terms to mean that relative URIs are a
      subset of URIs rather than a method of referencing URIs, this
      specification simply refers to them as relative references.

This suggests that `abc` is actually not a URI, but a relative reference.
While draft 6 [0] does distinguish between `uri` and `uri-reference`, draft 4 [1] only mentions `uri`, which makes me wonder if the authors may have meant that a `uri` in draft 4 could also be a relative reference. 

I would argue that keeping the lax interpretation of URI, as in version 1.5.1 and previous, is "more correct". It also certainly prevents the sort of backward incompability we have experienced in our software when trying to upgrade to version 1.6.1

[0]: Validation draft 6: https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-8.3.6
[1]: Validation draft 4: https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-7.3.6

lmontrieux:
Allow relative references as URIs in draft 4 validation.
Fixes #132

### DIFF
--- a/core/src/main/java/org/everit/json/schema/internal/URIV4FormatValidator.java
+++ b/core/src/main/java/org/everit/json/schema/internal/URIV4FormatValidator.java
@@ -1,0 +1,27 @@
+package org.everit.json.schema.internal;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Optional;
+import org.everit.json.schema.FormatValidator;
+
+public class URIV4FormatValidator implements FormatValidator {
+
+    @Override
+    public Optional<String> validate(final String subject) {
+        try {
+            URI uri = new URI(subject);
+            return Optional.empty();
+        } catch (URISyntaxException | NullPointerException e) {
+            return failure(subject);
+        }
+    }
+
+    protected Optional<String> failure(String subject) {
+        return Optional.of(String.format("[%s] is not a valid URI", subject));
+    }
+
+    @Override public String formatName() {
+        return "uri";
+    }
+}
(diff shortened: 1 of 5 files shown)


erosb:
Hello @lmontrieux , welcome back to the project. 

This issue took some time to investigate. It seems before version 1.6.0 we didn't run the format.json testcases for verifying compatibility, so that's why some strings passed the URI validation which shouldn't. "abc" doesn't seem to be valid against the testsuite of [any json-schema draft specifications](https://github.com/json-schema-org/JSON-Schema-Test-Suite) . Also, it was never valid [according to the history](https://github.com/json-schema-org/JSON-Schema-Test-Suite/commits/master/tests/draft4/optional/format.json). So it seems the current behavior conforms to the spec, and the versions before 1.6.0 did not.

On the other hand I try to take backward-compatibility seriously, and this change in 1.6.0 is definitely a BC break. So even if this change breaks spec conformance, it restores backward-compatibility, therefore the better thing to do in this situation is merging your PR.

It also worths noting that if someone needs strict draft-4 conformance, he can achieve it even after merging this PR by adding the current format validator as an explicit "uri" format validator by calling `SchemaLoader.builder().addFormatValidator(new URIFormatValidator());`.

Thank you for investigating and fixing this.

PR #137, opened 2018-01-20
V7 relative-json-pointer format

erosb:


core/src/main/java/org/everit/json/schema/internal/RelativeJsonPointerFormatValidator.java
```diff
@@ -0,0 +1,102 @@
...
+        private void parseUpwardsStepCount() throws ParseException {
+            if (!isDigit(curr())) {
+                fail();
+            }
+            if (pos == input.length()) {
+                return;
+            }
+            for (char current = curr(); isDigit(current) && pos < input.length() - 1; current = curr())
+                ;
+            pos--;
```
zgyorffi: Does it work for values that consists of only positive-integer and it has more than one digits e.g. '123'? In that case I think because of the step back the other parsing (parseJsonPointer) will try to process the ending digit.

This could be prevented by something like:
` if (!(pos == input.length() - 1)) {
pos--;
}`

zgyorffi:
One more thing: leading zero is not allowed before digits. Could you check and test this case too?

PR #144, opened 2018-02-10
If-then-else support

zgyorffi:


core/src/main/java/org/everit/json/schema/ConditionalSchemaValidatingVisitor.java
```diff
@@ -0,0 +1,61 @@
+package org.everit.json.schema;
+
+
+import java.util.Arrays;
+
+import static java.util.Objects.requireNonNull;
+
+public class ConditionalSchemaValidatingVisitor extends Visitor {
```
erosb: @zgyorffi If you want to utilize the visitor pattern, then please follow the current convention. The visitor design pattern aims to separate the traversal and the processing of the data it handles (the schema, in this case). In this context it means that in the `Visitor` superclass the following methods should also exist:

 * `visitIfSchema(Schema ifSchema)`
 * `visitThenSchema(Schema thenSchema)`
 * `visitElseSchema(Schema elseSchema)`

And the default `Visitor#visitConditionalSchema(ConditionalSchema)` method should call these in the above order. This is the separated _traversal_ step. Then you can override this 3 method in the `ConditionalSchemaValidatingVisitor` subclass to perform the validation. The class can have some mutable fields (eg. you will want to store the `ValidationException` of the `IfSchema`, if it exists), it won't hurt thread safety, because from the caller side there is only a local reference to the `ConditionalSchemaValidatingVisitor` instance.

The reason why I ask for this change is that this current implementation _seemingly_ follows the existing visitor structure, while in fact it doesn't.

core/src/main/java/org/everit/json/schema/ConditionalSchemaValidatingVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    private void visitThenSchema(ConditionalSchema conditionalSchema) {
+        if (conditionalSchema.getThenSchema().isPresent()) {
+            ValidationException thenSchemaException = owner.getFailureOfSchema(conditionalSchema.getThenSchema().get(), subject);
+            if (thenSchemaException != null) {
+                owner.failure(new ValidationException(conditionalSchema,
+                        new StringBuilder(new StringBuilder("#")),
+                        "Data is invalid for schema of \"then\" ",
```
erosb: Please change it to `"input is invalid against the \"then\" schema"`

core/src/main/java/org/everit/json/schema/ConditionalSchemaValidatingVisitor.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    private void visitElseSchema(ConditionalSchema conditionalSchema, ValidationException ifSchemaException) {
+        if (conditionalSchema.getElseSchema().isPresent()) {
+            ValidationException elseSchemaException = owner.getFailureOfSchema(conditionalSchema.getElseSchema().get(), subject);
+            if (elseSchemaException != null) {
+                owner.failure(new ValidationException(conditionalSchema,
+                        new StringBuilder(new StringBuilder("#")),
+                        "Data is invalid for schema of both \"if\" and \"else\" ",
```
erosb: Please change it to `"input is invalid against both the \"if\" and \"else\" schema"`

core/src/test/java/org/everit/json/schema/ConditionalSchemaTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Test
+    public void onlyElseSuccess() {
+        ConditionalSchema.builder().elseSchema(MAX_LENGTH_STRING_SCHEMA).build().validate("foo");
+    }
+
+    // if-then
+
+    @Test
+    public void ifSubschemaSuccessThenSubschemaFailure() {
```
erosb: If you pass a `Schema.Builder` to the `failureOf()` then it will call both `buildWithLocation()` and also set an expected pointer. You also don't have to explicitly pass the `expectedViolatedSchema` if it is the same as the root schema (it matters only when we check sub-schema failures). So this test can be written as 

```
ConditionalSchema.Builder subject = ConditionalSchema.builder().ifSchema(MAX_LENGTH_STRING_SCHEMA).thenSchema(PATTERN_STRING_SCHEMA);
        TestSupport.failureOf(subject)
                .expectedKeyword("then")
                .input("bar")
                .expect();
```

But if you prefer being more explicit then you can leave it as-is.

Issue #147, opened 2018-02-26
exponential runtime for pattern matching with Java's builtin PCRE regex engine


rkeytacked:
In this test case the JSON schema validation will end up eating all your CPU power for the next million years, without terminating.
- `schema.json`:
```
{"type":"object", "properties": {"prop": {"pattern": "a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?a?aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}}}
```
- `subject-valid.json`
```
{"prop": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"}
```

The reason behind is exponentially bad runtime behaviour of all PCRE regex engines.

You should consider using RE2 (https://github.com/google/re2j) instead, because this issue creates an easy denial-of-service-attack to all APIs using JSON schema validation for client-defined schemas.

rkeytacked:
As seen in the PR using `com.google.re2j.Pattern` instead of `java.util.regex.Pattern` fixes the exponential runtime.

But https://github.com/google/re2j is BSD-3 licensed. Is this compatible with this project's Apache 2 License?

erosb:
@rkeytacked in case you plan to upgrade to version 1.9.0 please keep in mind that we don't use re2j by default, because it caused some backward-compatibility problems for some users. You can still turn it on manually with a [one-line change](https://github.com/everit-org/json-schema#regexp-implementations).

rkeytacked:
Thanks @erosb for the info.
(And regarding the spec compliance of the regexes I'm still trying to figure out how to influence the community behind JSON Schema Spec in a way that we can remove this "vulnerability by design" in future versions…)

PR #148, opened 2018-02-26
test case for issue #147

rkeytacked:
- regular expression that takes ages to validate using PCRE libs for pattern matching


rkeytacked:
As expected the test case let's Travis CI run until timeout.

The quick idea to just replace `java.util.regex.Pattern` by `com.google.re2j.Pattern` fixes the exponential runtime, but raises a problem with the `equalsVerifier` tests.

rkeytacked:
Just saw that https://github.com/google/re2j is BSD-3 licensed. Is this compatible with this project's Apache 2 License?

rkeytacked:
ok, last commit fixes the `equalsVerifier` thingy, but now the code base is not backward compatible?

> Current codebase is not backward compatible (BACKWARD_COMPATIBLE_USER) with version <1.5.1>. Compatibility type has been detected as <NON_BACKWARD_COMPATIBLE>

erosb:
Hello @rkeytacked , 

thank you for raising this PR.

The incompatibility with the previous versions is that `StringSchema#getPattern()` returned `java.util.regex.Pattern` before your change, but after the change it returns `com.google.re2j.Pattern`.

I will think about options to resolve this. It would be good to merge the PR, it is an important improvement.

rkeytacked:
Hi @erosb , maybe it's possible to have a major version number change – like 1.6.0?

erosb:
Hello @rkeytacked ,

1.6.0 is already released, actually 1.7.0 is also out and 1.8.0 is coming in a few days (newer versions are available on JitPack, not maven central).

Minor versions should be backwards-compatible, so only 2.0 would solve the problem, but I absolutely don't think it is time for the next major release.

I could resolve the backward-compatibility problem locally by converting from/to `java.util.regex.Pattern` in all public methods. I merge your PR then I'm going to resolve the build failure in the master branch.

PR #149, opened 2018-02-26
If-then-else support: schema loading

zgyorffi:


core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -373,6 +390,8 @@ public SchemaLoader(SchemaLoaderBuilder builder) {
             builder = buildEnumSchema();
         } else if (ls.schemaJson().containsKey("const") && (config.specVersion != DRAFT_4)) {
             builder = buildConstSchema();
+        } else if (config.specVersion.compareTo(DRAFT_6) > 0 && schemaHasAnyOf(CONDITIONAL_SCHEMA_KEYWORDS)) {
```
erosb: Can we move it into `#sniffSchemaByProps()`  somehow?

core/src/test/java/org/everit/json/schema/loader/SchemaLoaderTest.java
```diff
@@ -140,6 +142,30 @@ public void integerSchema() {
...
+        Assert.assertFalse(actual.getElseSchema().isPresent());
+    }
+
+    @Test
+    public void conditionalSchemaElse() {
+        ConditionalSchema actual = (ConditionalSchema) loadAsV7(get("conditionalSchemaElse"));
+        Assert.assertFalse(actual.getIfSchema().isPresent());
+        Assert.assertFalse(actual.getThenSchema().isPresent());
+        Assert.assertTrue(actual.getElseSchema().isPresent());
+    }
```
erosb: Also it would be good to have tests for
 - checking that the if-then-else schema is loaded only in DRAFT_7 mode
 - the case where a subschema is a boolean (eg, `{"if" : true }` )
 - a case where all 3 keywords are present

core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -363,6 +366,20 @@ public SchemaLoader(SchemaLoaderBuilder builder) {
         return builder;
     }
 
+    private ConditionalSchema.Builder buildConditionalSchema() {
+        ConditionalSchema.Builder builder = ConditionalSchema.builder();
+        if (ls.schemaJson().containsKey("if")) {
+            builder.ifSchema(loadChild(ls.schemaJson().require("if").requireObject()).build());
+        }
```
erosb: The schema is not always an object, it can also be a boolean. so please don't call requireObject() here.
Also, instead of an explicit `if` followed by a `schemaJson().require("if")` call can be replaced by a `maybe("if")` call which returns an`Optional<JsonValue>`. So eg. line 371 - 373 can be replaced with something like:

`ls.schemaJson().maybe("if").map(this::loadChild).ifPresent(builder::ifSchema)`

The same goes for handling `then` and `else`.

PR #151, opened 2018-02-27
fix NPE when calling getPattern()

rkeytacked:
- fix NPE for `null` pattern when calling `getPattern()`
- increasing code coverage a tiny bit

related to issue #147


erosb:
Thanks @rkeytacked !

PR #153, opened 2018-03-01
Fix issue where loading the v6 schema overwrites all the custom validators

hansmire:
This request is to fix issue 152. The problem is the custom format validators get overwritten when you create the referenced schemas and you are using version 6 of json schema. The fix adds the standard format validators to the reference schema if they don't exist already. It does not to overwrite the validators copied from the parent. 

I hacked in a change to the IssueTest to allow me to add custom format validators. I think that the IssueTest probably needs to allow a SchemaBuilder to be passed in, but I did not want to make a proposal on how to do that. Let me know if there is a better way to add a custom format validator to IssueTest.

Also I made a change to always shut down the jetty service in IssueTest even if there is a failure.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/loader/LoadingState.java
+++ b/core/src/main/java/org/everit/json/schema/loader/LoadingState.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


erosb:
Hello @hansmire , thank you for submitting this PR.

Since there was a HUGE branch merged today into `master`, I didn't want to spend time on resolving conflicts, instead I copied most of your changes to the master branch.

So technically I close this PR, but I manually merged your changes, and included in the just now released version 1.8.0.

PR #166, opened 2018-04-15
Issue 160

zgyorffi:


core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -122,11 +122,23 @@ public SchemaLoaderBuilder draftV7Support() {
 
         private void setSpecVersion(SpecificationVersion specVersion) {
             this.specVersion = specVersion;
-            specVersion.defaultFormatValidators().forEach(this::addFormatValidator);
-            //            this.formatValidators = new HashMap<>(specVersion.defaultFormatValidators());
+        }
+
+        private Optional<SpecificationVersion> specVersionInMetaSchema() {
```
erosb: I think it should be called `specVersionInSchema()`

core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
```diff
@@ -122,11 +122,23 @@ public SchemaLoaderBuilder draftV7Support() {
...
+            Optional<SpecificationVersion> specVersion = Optional.empty();
+            if (schemaJson instanceof Map) {
+                Map<String, Object> schemaObj = (Map<String, Object>) schemaJson;
+                Object schemaValue = schemaObj.get("$schema");
+                if (schemaValue != null) {
+                    specVersion = Optional.of(SpecificationVersion.getByMetaSchemaUrl((String) schemaValue));
+                }
```
erosb: The above 4 lines could be simplified to `specVersion = Optional.ofNullable(schemaObj.getString("$schema")).map((SpecificationVersion::getByMetaSchemaUrl))`

zgyorffi:
The root cause of the problem is that default format validators are set at the point when the specversion is also set explicitly by the builder (draftV6Support,...). If the specversion is defined by $schema the format validators are not set accordingly because it is checked at a later point (in the constructor of SchemaLoader).

core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
zgyorffi: Small modification: schemaObj.get("$schema") instead of getString and a cast to String.

PR #183, opened 2018-05-30
Fix #182

### LINKED ISSUE #182: Json pointer references with the same name in different schema files are not resolved correctly
When using json pointers with the same name in different schema files, they all resolve to the schema that is pointed to by the first one that is resolved.

We have a schema that refers to two remote schema files.
Both of these files happen to have a `definition` with the same name, and both files refer to their own definition with a local json pointer (one that starts with `#`).

Expected behavior: Both references are resolved to the definition in their own file
Actual behavior: Both references are resolved to the definition that is resolved first

vierbergenlars:


erosb:
Hello @vierbergenlars , thank you for your contribution, much appreciated.

I merged your PR manually and improved the testcase a bit in bc51361f0b7d0f11929c1ec8e7d38d2e2ed9ed6b .

PR #190, opened 2018-06-19
Replaced the org.json JSON parser with jackson JSON parser

ramuramaiah:
Replaced the org.json JSON parser with jackson JSON parser


ramuramaiah:
erosb,
I have created a pull request.
The change is the following
"Replaced the org.json JSON parser with jackson JSON parser"

ramuramaiah:
Hi,
I have modified the original source code of everit JSON schema parser to make use of jackson as the underlying JSON schema parser instead of org.json due to the business unfriendly license of org.json JSON parser.
I have made sure the following

The original design and the interfaces remains the same. The modification is only minimal.
All tests pass.
Let me know if I can share the changes and anyone is interested to make use of the same.
I am not sure about the copyrights etc. Hence I don't want to create a new GitHub project.

erosb:
Hello @ramuramaiah , the library is licensed under apache 2 license, which is a very permissive license. You are free to re-distribute the library and/or use a modified version in a commercial closed-source project.

The PR is not likely to be merged as-is.

PR #192, opened 2018-06-29
Fix Maven dependency's conflicts

snazarenko67:
Hello!

There were maven dependency conflicts for artifacts:
- commons-collections
- commons-logging
- hamcrest-core

They were resolved in pom-file.
Thanks.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -154,12 +154,29 @@
             <groupId>commons-validator</groupId>
             <artifactId>commons-validator</artifactId>
             <version>1.6</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>commons-beanutils</artifactId>
+                    <groupId>commons-beanutils</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>commons-beanutils</groupId>
+            <artifactId>commons-beanutils</artifactId>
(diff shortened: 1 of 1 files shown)


erosb:
Hello!

I'm not sure if I understand the problem solved by this PR, and/or how does it make the library more useful. Can you please give some context?

snazarenko67:
Hello!

Problem exists in children maven dependencies of commons-validator, commons-beanutils, mockito-core and junit (it's visible in jetbrains as dependency conflicts of different versions artifacts).

commons-validator 1.6 uses: commons-beanutils 1.9.2, commons-logging 1.2 and commons-collections 3.2.2, but commons-beanutils 1.9.2 uses commons-logging 1.1.1 and commons-collections 3.2.1.

On this resason, the dependency commons-validator 1.6 to commons-beanutils 1.9.2 was excluded in pom, and the dependency commons-beanutils 1.9.3 was added (it uses commons-logging 1.2 and commons-collections 3.2.2).

mockito-core 1.10.19 references to hamcrest-core 1.1, but junit 4.12 references to hamcrest-core 1.3.
The references from mockito-core to hamcrest-core was excluded in pom.

So, we resolve conflicts and know exact versions of all maven artifact.

That's all.

erosb:
Aha, I see now, thanks!

PR #193, opened 2018-07-02
Remove Guava and Refactor format validators

snazarenko67:
Hello!

In my project I use the json-schema-validator in a distributed enterprise service bus (without maven support). Guava package looks too big for it.
The using of guava was limited to _com.google.common.net.InternetDomainName_ and _com.google.common.collect.ImmutableList_.

_ImmutableList_ is replaced to _java.util.Collections.unmodifiableList_.
_InternetDomainName_ is replaced to _org.apache.commons.validator.routines.DomainValidator_.

Now Guava is used in test scope only.

Besides this, I looked at code of format validators and made the following changes:
- I removed NPE exception handling;
- IPv4/IPv6 validators use _org.apache.commons.validator.routines.InetAddressValidator_, I remove superclass IPAddressValidator;
- I removed redundant inner class Delegate from the DateTimeFormatValidator;
- I removed duplicate code from enum _org.everit.json.schema.loader.SpecificationVersion_, and I used the format name from validator instance instead of hardcoding (a potential error excluded).

Thanks.


core/src/main/java/org/everit/json/schema/internal/TimeFormatValidator.java
```diff
@@ -1,16 +1,15 @@
...
 import java.time.format.DateTimeFormatterBuilder;
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * Implementation of the "time" format value.
  */
 public class TimeFormatValidator extends TemporalFormatValidator {
-    private static final List<String> FORMATS_ACCEPTED = ImmutableList.of(
+    private static final List<String> FORMATS_ACCEPTED = Arrays.asList(
```
agebhar1: it should be `Collections.singletonList(...`

snazarenko67:
```java
private static final List<String> FORMATS_ACCEPTED = Arrays.asList(
            "HH:mm:ssZ", "HH:mm:ss.[0-9]{1,9}Z, HH:mm:ss[+-]HH:mm, HH:mm:ss.[0-9]{1,9}[+-]HH:mm"
    );
```
The array has more than one elements.

snazarenko67: This logic (Collections.unmodifiableList) moved to TemporalFormatValidator:

```java
    TemporalFormatValidator(DateTimeFormatter formatter, List<String> formatsAccepted) {
        this.formatter = requireNonNull(formatter, "formatter cannot be null");
        this.formatsAccepted = Collections.unmodifiableList(formatsAccepted);
    }
```

agebhar1: :+1:

core/src/test/java/org/everit/json/schema/loader/SchemaLoaderTest.java
```diff
@@ -687,14 +655,18 @@ public void unknownMetaSchemaException() {
     }
 
     @Test
-    public void syntheticAllOf() {
```
erosb: @snazarenko67 can you please restore this test?

core/src/test/java/org/everit/json/schema/loader/SchemaLoaderTest.java
```diff
@@ -1,53 +1,7 @@
...
-import org.everit.json.schema.TestSupport;
-import org.everit.json.schema.TrueSchema;
-import org.everit.json.schema.internal.DateTimeFormatValidator;
-import org.everit.json.schema.internal.EmailFormatValidator;
-import org.everit.json.schema.internal.HostnameFormatValidator;
-import org.everit.json.schema.internal.IPV4Validator;
-import org.everit.json.schema.internal.IPV6Validator;
-import org.everit.json.schema.internal.URIV4FormatValidator;
+import org.everit.json.schema.*;
+import org.everit.json.schema.internal.*;
```
erosb: I would more prefer avoiding import-star declarations everywhere, this has been a convention for the libary since its origins.

core/src/main/java/org/everit/json/schema/loader/SpecificationVersion.java
```diff
@@ -125,44 +112,39 @@ static SpecificationVersion getByMetaSchemaUrl(String metaSchemaUrl) {
...
+            new DateFormatValidator(),
+            new URIFormatValidator(false),
+            new TimeFormatValidator(),
+            new RegexFormatValidator(),
+            new RelativeJsonPointerFormatValidator()
+    );
+
+    private static Map<String, FormatValidator> formatValidators(Map<String, FormatValidator> parent, FormatValidator... validators) {
+        Map<String, FormatValidator> validatorMap = parent == null ? new HashMap<>() : new HashMap<>(parent);
+        for(FormatValidator validator : validators)
```
erosb: Can you please add the brackets here?

erosb:
Hello @snazarenko67 are you planning to address the comments on the PR?

core/src/test/java/org/everit/json/schema/loader/SchemaLoaderTest.java
snazarenko67: Hi!
I will look at all the comments and answer ASAP.

snazarenko67:
Hello!
All comments processed.

core/src/main/java/org/everit/json/schema/internal/TemporalFormatValidator.java
```diff
@@ -26,7 +26,7 @@
 
     TemporalFormatValidator(DateTimeFormatter formatter, List<String> formatsAccepted) {
         this.formatter = requireNonNull(formatter, "formatter cannot be null");
-        this.formatsAccepted = new ArrayList<>(formatsAccepted);
+        this.formatsAccepted = Collections.unmodifiableList(formatsAccepted);
```
erosb: `Collections.unmodifiableList()` just returns an immutable view of the original array, but doesn't copy it, therefore here if the list gets mutated on the caller side, then the change will be reflected in `this.formatsAccepted` . The proper defensive way of this copy without guava is `this.formatsAccepted = unmodifiableList(new ArrayList<>(formatsAccepted))`.

core/src/main/java/org/everit/json/schema/internal/DateTimeFormatValidator.java
```diff
@@ -27,23 +26,18 @@
         }
     }
 
-    private static final List<String> FORMATS_ACCEPTED = ImmutableList.of(
+    private static final List<String> FORMATS_ACCEPTED = Arrays.asList(
```
erosb: `Arrays.asList()` returns a fixed-size but not immutable list. It would be better to wrap this into a `Collections.unmodifiableList()` too.

erosb:
It looks good, merged. Thank you for your contribution @snazarenko67 and also for the reviews @agebhar1 @zgyorffi !

Issue #194, opened 2018-07-12
NPE in org.everit.json.schema.ValidationException#hashCode


EXPEaneparaczki:
Hello! It's possible to get NPE-s when trying to add ValidationException-s to any hash based collection. This  happened to us using the library. Solution is to make org.everit.json.schema.ValidationException#hashCode exception safe. In our concrete case the 'keyword' field was the one with null.

EXPEaneparaczki:
Hi! thanks for the quick response. In my opinion there is no need for that. The root cause is the possibility of the hashCode method throwing NPE-s which generally shouldn't be the case.

PR #195, opened 2018-07-12
fixing #194 by making ValidationException#hashCode() null-tolerant

erosb:


davideicardi:
Can you please provide an official release with this fix? Or there is an unstable release somewhere?

NullPointerException breaks my log4j code and I cannot log any errors of this kind.

thanks

erosb:
Hello @davideicardi , the cool thing about JitPack is that it lets you use git commits as version numbers, so if you set the current HEAD rev as version number, then you will get the fix:

```        <dependency>
            <version>63c3f139a5fdb2b8837da56b2d128b483e0d1943</version>
        </dependency>```

PR #198, opened 2018-07-16
 add support for other integer types derived from java.math.Number

EXPEaneparaczki:


core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java
```diff
@@ -3,9 +3,18 @@
 import static java.lang.String.format;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.HashSet;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.google.common.collect.Sets;
```
erosb: Currently we don't use guava, this import doesn't compile on CI ( https://travis-ci.org/everit-org/json-schema/builds/404507721 )

core/src/test/java/org/everit/json/schema/NumberSchemaTest.java
```diff
@@ -253,5 +258,26 @@ public void requiresInteger_nonNullable() {
...
+        NumberSchema subject = NumberSchema.builder().build();
+        
+        tryIntegerTypes(subject);
+    }
+
+    private void tryIntegerTypes(NumberSchema subject) {
+        subject.validate(BigInteger.valueOf(123123123123123L));
+        subject.validate(BigDecimal.valueOf(123123123123123L));
+        subject.validate(new AtomicInteger(123));
+        subject.validate(new AtomicLong(123L));
```
erosb: It would make sense to differentiate between integral types (`BigInteger`, `AtomicInteger` and `AtomicLong`), and number types (`BigDecimal`). JSON schema has both "number" and "integer" keywords, and these are both handles by `NumberSchema`. `BigDecimal`  is definitely a number but often not an integer.

aneparaczki: my bad, BigDecimal is in the wrong place

core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java
aneparaczki: right

aneparaczki:
the json lib works so that JSONObjects made from POJOs that have ie. a BigInteger property, will have this BigInteger in it's value map, in which case the above modified code gets called on this exact BigInteger right here: https://github.com/EXPEaneparaczki/json-schema/blob/b35d606cb39c4a3ca96c1f9d3bab485517d04f84/core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java#L35. It's the subject. It's needed because we might have numbers that doesn't fit long, so we should support POJOs with these long numbers.

erosb:
Looks good, merged, thank you for your contribution @EXPEaneparaczki .

EXPEaneparaczki:
Thank you!

Issue #227, opened 2018-10-16
Does validation happen for nested entities referred too?


indrajra:
It appears that only the root schema is validated. The nested child entity schema isn't. Is this expected or is there a setting that I am missing to enforce this? I'd like to validate against a nested child entity schema too, as long as they are internal (not referring to external world).

erosb:
Hello, this library validates json objects and not entities, but anyway, it definitely does validate child objects.

 Please raise a PR with a failing test [as described in the contribution guide](https://github.com/everit-org/json-schema/blob/master/CONTRIBUTING.md) to demonstrate your problem.

indrajra:
Hi @erosb - I've created PR https://github.com/everit-org/json-schema/pull/228 for this. Take a look. Sorry, I had used object and entity in a synonymous way. Thanks for the correction.

erosb:
Hello, I can see that this  reference causes the SchemaException: ` "$ref": "subdir/color.json#/definitions/name"`

It fails because it is a relative URI, and the loader needs an absolute base URI to resolve it against. It is documented as [resolution scopes](https://github.com/everit-org/json-schema#resolution-scopes).

indrajra:
Well, if I set the resolution scope as suggested, it leads to a connection refused error. Is it that this library expects a http server even for local relative references?

indrajra:
Once I setup a http server (for trial), I could see the child schemas was getting validated. Thanks @erosb for your guidance on this. I'm marking this ticket closed and withdraw my PR.

indrajra:
Closing this issue and PR associated. I served the child schema through a http server and thereafter the validation happened on the child object as expected.

PR #228, opened 2018-10-18
Issue #227-Validate child objects

indrajra:
Hi there, attached is a test case that fails with 
`java.io.UncheckedIOException: java.net.MalformedURLException: no protocol: subdir/color.json
	at org.everit.json.schema.loader.internal.DefaultSchemaClient.get(DefaultSchemaClient.java:20)
	at org.everit.json.schema.loader.JsonPointerEvaluator.executeWith(JsonPointerEvaluator.java:76)`

The need is to have multiple schemas defining its own objects; one object defined in one separate schema file. I want to be able to refer to the child objects and validate them, allowing users to pass entire payload (parent + child objects data) in a nested form.

I expect all the schemas to be available locally and very few in remote. For all local schemas, I'd like to have strict validation and for those in remote, on a case-to-case basis.

### DIFF
--- a/tests/src/test/resources/org/everit/json/schema/issues/issue227/remotes.subdir/color.json
+++ b/tests/src/test/resources/org/everit/json/schema/issues/issue227/remotes.subdir/color.json
@@ -1,5 +1,5 @@
 {
-  "id":"http://localhost:1234/color.json#",
+  "id":"http://localhost:1234/subdir/color.json#",
   "definitions" : {
     "name" : {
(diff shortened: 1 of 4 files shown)


indrajra:
Once the resolution scope was setup, the validation happened just fine. I'm withdrawing the additional test/issue I submitted.

indrajra:
Once the resolution scope was set as advised in the issue, the validation on the child object happened, which is fine and expected.

Issue #189, opened 2018-06-19
jackson as the JSON parser instead of org.json


ramuramaiah:
Hi,
I have modified the original source code of everit JSON schema parser to make use of jackson as the underlying JSON schema parser instead of org.json due to the business unfriendly license of org.json JSON parser.
I have made sure the following
* The original design and the interfaces remains the same. The modification is only minimal.
* All tests pass.

Let me know if I can share the changes and anyone is interested to make use of the same.
I am not sure about the copyrights etc. Hence I don't want to create a new GitHub project.

erosb:
Hello @ramuramaiah , thank you for your efforts. This sounds promising, though I doubt it doesn't break any of the method signatures. Please create a fork of this project and share your changes in the form of a pull request. If you are not familiar with this workflow please refer to [the GitHub documentation](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) .

Making the json implementation replaceable has been under consideration for quite some while, but for me it doesn't look simple to replace the data model of the library in a backward-compatible way (which is a top-priority for the project). I'm curious about what you figured out.

Thanks.

ramuramaiah:
erosb,
Do I need some permissions to create a pull request for this project? The "create pull request" tab is disabled for me. I did login with my GitHub account.

erosb:
You have to go to the [GH page of your fork](https://github.com/ramuramaiah/json-schema) and find the Pull Request button there.

ramuramaiah:
erosb,
I have created a new pull request and you should be able to see it.
Following is the heading of the pull request
>>
Replaced the org.json JSON parser with jackson JSON parser #190
<<
Please review and let me know your opinion.

Yobikap:
@ramuramaiah Is it possible to create a fork and implement your changes into that fork?

I also need this functionality as for now it is giving me a lot of headache to let the code work with existing POJO's and with Jackson

erosb:
@Yobikap if you need jackson then TBH I would look for an other json schema library. There [are a few options nowadays](https://json-schema.org/implementations.html#validator-java).

Issue #37, opened 2016-06-17
org json vs google gson


mspiegel:
I don't mean to cause controversy but what's the merit of using org.json rather than gson? The org.json license is not accepted within most corporations which prevents the use of otherwise great projects. See https://dzone.com/articles/jsonorg-license-literally-says. Is org.json significantly faster than gson? Is it more popular than gson? I don't mind submitting a pull request that would move it to gson, I don't think that would be a lot of work.

erosb:
It isn't really a good question here, since this project is not the org.json library itself but a json-schema implementation on the top of it. Maybe better to ask at this repo: https://github.com/stleary/JSON-java

Anyway, IMO
> Is org.json significantly faster than gson?

GSON is a data mapping library, involving a massive use of reflection (you map JSON to arbitary java objects), while org.json is a generic JSON representation (it parses json strings itno JSONObject and JSONArray instances), and a DSL for generating JSON text (with JSONWriter). So the scope of the two libraries are slightly different. Due to GSON being more reflection-heavy, I think it is probably significantly slower than  org.json, but I've never measured it.

> Is it more popular than gson?

Not at all. I can see GSON and Jackson being the most popular solutions for managing JSON data in the java world.

> I don't mind submitting a pull request that would move it to gson, I don't think that would be a lot of work.

GSON is, as far as I'm concerned, is a data mapping library, so the schema is specified by the java classes you serialize JSON from/to. Supporting GSON with a json-schema library just doesn't make sense for me, but maybe I'm wrong.

mspiegel:
https://github.com/stleary/JSON-java has a license with the clause "The Software shall be used for Good, not Evil." It's not productive to post an issue with that repository because Douglas Crockford added that clause to his project and he is not interested in removing it. I don't really have a problem with that part of the license. It's just I can't use that library for my work projects because the lawyers won't let me.

Because this library uses that library as a dependency, I can't use this library either for work. Sorry it's been about half a year since I've used json libraries in Java I forgot that GSON was more than just a parsing library. It's a shame there are not more small json parsing libraries for Java.

What if this library used only the json parsing functionality from GSON? I saw there is a parser in GSON. I don't mind doing the work and submitting a pull request. But if there's not an interest in moving away from the org.json parser then I won't do unnecessary work.

erosb:
It is a petty that the license of the org.json library causes you trouble.

I think if we want to provide a way in this library to avoid the use of org.json, then:
[...]

PR #229, opened 2018-10-22
Refactor JSON API into own Module

BjoernAkAManf:
This PR should not be merged as is and is simply a basic design that could be leveraged in the future. I did not document alot of the new functions, because most of them simply wrap the already existing methods / refactor them into new interfaces. This would allow to swap between implementations easily.

This PR is directly in contrast to #190 due to the nature of this PR. While i agree, that supporting Jackson is a good thing, i would kindly disagree, to simply switch the whole code base into a new model that is not backed by any standard (Obligatory [XKCD](https://xkcd.com/927/)).

See #37 #189 for more context to this issue.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -143,11 +143,13 @@
         </plugins>
     </build>
     <dependencies>
+        <!-- TODO: Move into own artifact-->
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
             <version>20180130</version>
         </dependency>
+        <!-- -->
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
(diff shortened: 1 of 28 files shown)


erosb:
Hello @BjoernAkAManf , 

Thank you for putting effort into this PR. This is the 3rd attempt (see #190 and #220 ) to abstract the library away from org.json. We should spend time on taking the best of all three and have a solid solution for the problem.

BjoernAkAManf:
I most certainly agree. For that we should define goals first. Such as backwards compatibility as a priority or not. I noticed many problems with the latter due to many public methods exposed, that some client might depend on. So it would be good idea to limit them in the future, define a deprecation strategy and remove / hide such methods.

PR #235, opened 2018-10-30
Issue 234 updated with json files

cannondt:
I hope I have followed the contributing guidelines correctly and this PR is correct, this is my first time doing this :)

### DIFF
--- a/tests/src/test/resources/org/everit/json/schema/issues/issue234/schema.json
+++ b/tests/src/test/resources/org/everit/json/schema/issues/issue234/schema.json
@@ -1,0 +1,140 @@
+{
+  "swagger" : "2.0",
+  "info" : {
+    "description" : "This is a simple API",
+    "version" : "1.0.0",
+    "title" : "Simple Inventory API",
+    "contact" : {
+      "email" : "you@your-company.com"
+    },
+    "license" : {
+      "name" : "Apache 2.0",
+      "url" : "http://www.apache.org/licenses/LICENSE-2.0.html"
+    }
+  },
+  "host" : "virtserver.swaggerhub.com",
+  "basePath" : "/intellectdesign/SchemaValidation/1.0.0",
+  "tags" : [ {
+    "name" : "admins",
+    "description" : "Secured Admin-only calls"
+  }, {
+    "name" : "developers",
+    "description" : "Operations available to regular developers"
+  } ],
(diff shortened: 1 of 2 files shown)


erosb:
Hi! This is a swagger json, with nested json schemas. This library isn't aware of swagger and does not extract nested schemas. Since there is no json schema keyword at the root level, this will be parsed as an empty json schema (`{}`), which accepts any json (so you don't get the validation error you expect).

cannondt:
Hi,
That is what I was referring to in the initial issue question. So how can I achieve this? I am forced to stick to swagger, yet I have to do a schema level validation. Can you help out with an alternative?

erosb:
There are multiple ways to work it around:

1) the cleaner one is be externalizing the json schemas with $ref : for example you can put the schema into `InventoryItem.json` alongside with your `swagger.json` , and instead of ` "$ref" : "#/definitions/InventoryItem"` you could do  `"$ref" : "./InventoryItem.json"`. Then `InventoryItem.json` would be a bare json schema, you could directly use it with this library, and your swagger json would still work (though you might hit some implementation-specific issues with loading the external json file).

2) you can just parse the entire swagger json, then query the json schema from it: `JSONObject schemaJson = new JSONObject("your-swagger-json-content").getJSONObject("definitions").getJSONObject("InventoryItem");` . This is a bit dirty but works if you can't make any modification to the swagger json document.

cannondt:
Option 1 is cleaner but not workable at all our swagger is centrally used by too many dependent parties. 

In option 2, how will there be a json scheme keyword? Is there no java library that can do the extraction of schema from swagger? Just curious and I really appreciate all the help you are trying to provide !

erosb:
> how will there be a json scheme keyword?
Do you mean `$schema` ?

> Is there no java library that can do the extraction of schema from swagger?
I didn't do any research on that.

cannondt:
> Do you mean `$schema` ?

Yes, in the step where we parse swagger to get a JSON object, would there be the keyword that kicks off the processing by this library?

There seems no reliable way of getting a usable json schema from swagger, so I am going down the path of option 2 :)

cannondt:
Great, will try out your suggestions and let you know how I get on. Thanks a ton for all the patient help 👍

erosb:
Closing (I assume it has been resolved in the meanwhile).

PR #239, opened 2018-11-09
jsonschema-jdk6-backport

mindbender1:
Hi!

The following things have been done:
1) [streamsupport](https://github.com/stefan-zobel/streamsupport)
2) [retrolambda](https://github.com/luontola/retrolambda)
3) [ThreeTen backport](https://www.threeten.org/threetenbp/)
4) Slightly modified copy of UncheckedIOException has been added to the project
5) Replaced .forEach with normal for each loops
6) Uncommented japicmp-maven-plugin in jsonschema/core/pom.xml

It is currently configured to create 1.6 bytecode.
Look at the retrolambda plugin in json-schema/core/pom.xml to configure it for 1.7

Please compile a jdk6 and jdk7 compliant version and distribute it to the world :-)

### DIFF
--- a/core/src/main/java/org/everit/json/schema/SchemaException.java
+++ b/core/src/main/java/org/everit/json/schema/SchemaException.java
@@ -1,11 +1,11 @@
 package org.everit.json.schema;
 
-import java8.util.stream.Collectors;
 import java8.util.stream.StreamSupport;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java8.util.Objects.requireNonNull;
+import static java8.util.stream.Collectors.joining;
 
 import java.util.Collection;
 import java.util.List;
@@ -45,7 +45,7 @@ public class SchemaException extends RuntimeException {
     static String buildMessage(String formattedPointer, String actualTypeDescr, Collection<Class<?>> expectedTypes) {
         String fmtExpectedTypes = StreamSupport.stream(expectedTypes)
                 .map(Class::getSimpleName)
-                .collect(Collectors.joining(" or "));
+                .collect(joining(" or "));
         return  format("%s: expected type is one of %s, found: %s", formattedPointer,
                 fmtExpectedTypes,
                 actualTypeDescr);
@@ -57,7 +57,7 @@ public class SchemaException extends RuntimeException {
     }
 
     private static String joinClassNames(final List<Class<?>> expectedTypes) {
-        return StreamSupport.stream(expectedTypes).map(Class::getSimpleName).collect(Collectors.joining(", "));
+        return StreamSupport.stream(expectedTypes).map(Class::getSimpleName).collect(joining(", "));
     }
 
     private final String schemaLocation;
--- a/core/src/main/java/org/everit/json/schema/SchemaException.java
+++ b/core/src/main/java/org/everit/json/schema/SchemaException.java
@@ -1,9 +1,11 @@
 package org.everit.json.schema;
 
+import java8.util.stream.Collectors;
+import java8.util.stream.StreamSupport;
+
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
-import static java.util.Objects.requireNonNull;
-import static java.util.stream.Collectors.joining;
+import static java8.util.Objects.requireNonNull;
 
 import java.util.Collection;
 import java.util.List;
@@ -41,12 +43,13 @@ public class SchemaException extends RuntimeException {
     }
 
     static String buildMessage(String formattedPointer, String actualTypeDescr, Collection<Class<?>> expectedTypes) {
-        String fmtExpectedTypes = expectedTypes.stream()
+        String fmtExpectedTypes = StreamSupport.stream(expectedTypes)
                 .map(Class::getSimpleName)
-                .collect(joining(" or "));
-        return format("%s: expected type is one of %s, found: %s", formattedPointer,
+                .collect(Collectors.joining(" or "));
+        return  format("%s: expected type is one of %s, found: %s", formattedPointer,
                 fmtExpectedTypes,
                 actualTypeDescr);
+
     }
 
     private static String buildMessage(String pointer, Class<?> actualType, Collection<Class<?>> expectedTypes) {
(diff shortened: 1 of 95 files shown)


mindbender1:
for some reasons it has errors in travis ci....but it works when i compile it locally on intellij / windows 10 / jdk6

mindbender1:
after fixing the tests i have to admit that it works when i compile it with jdk6 and jdk8.

do you have a clue why the checks fail?

erosb:
Hello, I tried to compile your branch from the command line with `mvn clean install` and it gave the same output as CI.

mindbender1:
you are right. in commandline i get the same result. only in intellij it worked....

mindbender1:
ok, the only thing left is the static methods in interfaces i believe. i am not totally sure how to fix them.

 Do you have a clue?

erosb:
Hello,

I'm afraid there is no way to implement static default methods in java7. First, that means you have to change the API and implement these functionalities in a java7 way.

It also means that the java7 backport can't be done without changing the public API of the library and breaking backward-compatiblity, and we definitely won't do that. As a consequence, this PR can't be merged (actually even if you succeed to make all the tests pass, there is a following build step which checks backward compatibility with the previous release, and most probably that would fail). 

On the other hand it is a _fairly_ simple diff (compared to previous java6 backport attempts) and it seems there is still value in a java6-compatible version. So what we can do is releasing a separate version with this patch, with an artifactId different than `org.everit.json.schema`. There are two ways to do it:
 * you can release a version on JitPack (that's nearly zero-effort, since jitpack uses github), or maybe even deploy it to maven central on your own (this process requires a bit of effort). If you want to keep up working on this in the future (releasing jdk7-compliant versions for each release) then probably it does worth it to maintain the backport this way.
 * I can merge the patch into my own fork then publish a release with some maven coordinates other than the `com.github.everit-org.json-schema:org.everit.json.schema` . This may make more sense if you don't plan to keep the backport up-to-date with the java8 version, so we treat it as a one-time effort and release (which is still highly beneficial for java7 users).

Which one would you prefer?

core/src/main/java/org/everit/json/schema/SchemaException.java
```diff
@@ -1,9 +1,11 @@
 package org.everit.json.schema;
 
+import java8.util.stream.Collectors;
+import java8.util.stream.StreamSupport;
+
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
-import static java.util.Objects.requireNonNull;
-import static java.util.stream.Collectors.joining;
```
zgyorffi: Why don't you use static import for java8 library here as well like "import static java8.util.stream.Collectors.joining;" instead of importing the whole class?

core/src/main/java/org/everit/json/schema/combatibility/UncheckedIOException.java
```diff
@@ -0,0 +1,64 @@
+package org.everit.json.schema.combatibility;
+
+import java8.util.Objects;
+
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+
+public class UncheckedIOException  extends RuntimeException {
```
zgyorffi: One extra space before extends keyword.

mindbender1:
> Hello,
> 
> I'm afraid there is no way to implement static default methods in java7. First, that means you have to change the API and implement these functionalities in a java7 way.
> 
> It also means that the java7 backport can't be done without changing the public API of the library and breaking backward-compatiblity, and we definitely won't do that. As a consequence, this PR can't be merged (actually even if you succeed to make all the tests pass, there is a following build step which checks backward compatibility with the previous release, and most probably that would fail).
> 
> On the other hand it is a _fairly_ simple diff (compared to previous java6 backport attempts) and it seems there is still value in a java6-compatible version. So what we can do is releasing a separate version with this patch, with an artifactId different than `org.everit.json.schema`. There are two ways to do it:
> 
> * you can release a version on JitPack (that's nearly zero-effort, since jitpack uses github), or maybe even deploy it to maven central on your own (this process requires a bit of effort). If you want to keep up working on this in the future (releasing jdk7-compliant versions for each release) then probably it does worth it to maintain the backport this way.
> * I can merge the patch into my own fork then publish a release with some maven coordinates other than the `com.github.everit-org.json-schema:org.everit.json.schema` . This may make more sense if you don't plan to keep the backport up-to-date with the java8 version, so we treat it as a one-time effort and release (which is still highly beneficial for java7 users).
> 
> Which one would you prefer?

To be honest, i would prefer that you merge it in your own fork and publish a one time release one your own. To be honest, i was just interested in how to backport an application from java8 to java6 - what needs to be done, which frameworks are available, what challenges you face in doing that. 

If my small help results in a onetime java6/java7 compliant release by you i am really fine with that!

erosb:
Ok, I can put effort into the release process this weekend at soonest. Can you please address the PR comments of @zgyorffi in the meanhwile?
Thank you.

mindbender1:
done.

My 2 cents: 
static methods: just create an additional class which only includes the static method
default methods: maybe just add abstract classes in the middle and let the formatters extend those instead of the interface....for example

`public abstract class AFormatValidator implements FormatValidator {
      String formatName() {
        return "unnamed-format";
    }
}`

erosb:
> static methods: just create an additional class which only includes the static method
> default methods: maybe just add abstract classes in the middle and let the formatters extend those instead of the interface....for example

+1, that's a straigthforward way to go.

mindbender1:
Ok, so i continued the backporting. Currently all but one single JUnit test is working. i ignored to test if the build works at all. 
Apart from that it is running smoothly in my POC project (jdk 6).

It is SchemaLoaderTest.customFormat().
Does someone have an idea how to fix this?

erosb:
Hello @mindbender1 , I released your branch to maven central with coordinates [com.github.erosb:everit-json-schema-jdk6](https://search.maven.org/artifact/com.github.erosb/everit-json-schema-jdk6/1.9.2/bundle). 

I'm closing this branch, seems unnecessary to resolve the conflicts which appeared in the meanwhile. I will also update the readme to refer to this version.

Thank you for your contribution @mindbender1 , hats off ! :tophat:

mindbender1:
Great, thanks for the flowers!

Open sources rulez 👍

PR #242, opened 2018-11-13
VisitorListener - Add a SchemaVisitorListener to know which schemas are matching against the event

afranzi:
This PR aims to allow developers to add a SchemaVisitorListener to know which *schemas* are matching against the validated event.

This is useful when you have some fields that are *OneOf* or *AnyOf*.

I'm wondering if the Listener should be provided as an interface so developers can extend it.

Any ideas or suggestions?


afranzi:
I've just changed to an interface so any developer is free to provide their own implementation.
What do you think @erosb?

erosb:
Hello, it is still on my todo list to review this PR, and unfortunately it will stay there for a couple of days. Thank you for putting effort into it.

core/src/main/java/org/everit/json/schema/ValidatingVisitor.java
```diff
@@ -46,13 +48,18 @@ void visit(Schema schema) {
...
         if (subject != null && !VALIDATED_TYPES.stream().anyMatch(type -> type.isAssignableFrom(subject.getClass()))) {
             throw new IllegalArgumentException(format(TYPE_FAILURE_MSG, subject.getClass().getSimpleName()));
         }
         this.subject = subject;
         this.failureReporter = failureReporter;
         this.readWriteValidator = readWriteValidator;
+        this.schemaVisitorListener = schemaVisitorListener;
+    }
+
+    ValidatingVisitor(Object subject, ValidationFailureReporter failureReporter, ReadWriteValidator readWriteValidator) {
```
erosb: why do we need this constructor?

core/src/main/java/org/everit/json/schema/ValidatingVisitor.java
```diff
@@ -33,6 +33,8 @@ private static boolean isNull(Object obj) {
 
     protected Object subject;
 
+    private SchemaVisitorListener schemaVisitorListener;
```
erosb: can we make this `final`?

core/src/main/java/org/everit/json/schema/ValidatingVisitor.java
afranzi: To keep compatibility with previous versions and people not using the SchemaVisitorListener. Since there are no default arguments in java I was keeping it.

afranzi:
Cool! Thanks for review  😄 

> maybe it would be better to restrict this feature to combined ("anyOf"/"oneOf" and maybe if-then-else) schemas. I can see no reason to trigger it for other schema types, because if those match/don't match then it will result in a root-level pass/failure. So if we trigger these events for all schemas, then

That would imply reporting the schema to the listener only inside the `visitReferenceSchema`? 
https://github.com/everit-org/json-schema/blob/367d9733fbbd4b9f7d34e7ea552aa27af4e2c004/core/src/main/java/org/everit/json/schema/ValidatingVisitor.java#L119

> instead of only passing the schema itself, better to wrap it into a ValidationEvent (or Context?) object with a single member. This would be useful because if necessary then in the future we can add more members to the event object, without breaking interface implementations of existing users.

If I understand the idea would be to have different Events that could be reported by the SchemaVisitorListener, starting with the `subschemaMath` & `subschemaMismatch`? It that's the case, it LGTM!

> I would call the interface methods subschemaMatch(event) and subschemaMismatch(event)

LGTM!

core/src/main/java/org/everit/json/schema/ValidatingVisitor.java
afranzi: Cool! I like removing things!

erosb:
> That would imply reporting the schema to the listener only inside the visitReferenceSchema?

No, this change doesn't affect the `ReferenceSchema` (`"$ref"` is a separate concept, it is just often used together with anyOf/oneOf). The reporting should happen in [visitCombinedSchema](https://github.com/everit-org/json-schema/blob/367d9733fbbd4b9f7d34e7ea552aa27af4e2c004/core/src/main/java/org/everit/json/schema/ValidatingVisitor.java#L141) and in `ConditionalSchemaValidatingVisitor` .

> If I understand the idea would be to have different Events that could be reported by the SchemaVisitorListener, starting with the subschemaMath & subschemaMismatch?

Yes, sort of. Multiple `subschemaMismatch` / `Match` methods, either with different names and/or different parameter type (I'm not sure at the moment which one would be better).

afranzi:
Hi @erosb,
I applied your proposal of moving the VisitorListener to be more event oriented and added a new event for the ReferencedSchemas.

I've detected some strange behavior in the `refSchema3` for the *location* field.
As you can see in the example, one schema is referenced using the `$if` and the other one the path to the definitions. Then, the location of the one using `$id` is valid, but the other one is null.
Is that expected?

afranzi:
@erosb any update? what do you think about the use case that we have?

PR #248, opened 2018-12-10
Improved ValidationListener

erosb:
This PR is an improved attempt for #242 .

Here there are a couple of separate listener methods for intermediate schema validation results (matches and mismatches) which help resolving validation ambiguity for the caller. There are also a couple of event types for matches and mismatches so that for each specific event type the listener implementation can be informed about the details of the event in an event-type-specific manner. The events (`ValidationEvent` being the root of the hierarchy) also provide a JSON description of themselves.

The created event hierarchy looks like

 * ValidationEvent
   * SchemaReferredEvent
   * CombinedSchemaValidationEvent (abstract)
     * CombinedSchemaMatchEvent
     * CombinedSchemaMismatchEvent
   * ConditionalSchemaValidationEvent (abstract)
     * ConditionalSchemaMatchEvent
     * ConditionalSchemaMismatchEvent
     
     
Also a couple of things were recognized in the progress:
 * CombinedSchema didn't properly override #describePropertiesTo() to #toString() returned an empty json array
 * FalseSchema#INSTANCE was added (similar to TrueSchema#INSTANCE)
 * properly overriding TrueSchema#equals() so that it doesn't accept EmptySchema instances
 * fixed ValidationException#equals() to take the exception message into account

@afranzi please review the PR and/or try it out locally if it fits your needs, then if there is no major problem then we can proceed with releasing it. Thanks.

PS. @afranzi somehow I pushed your VisitorListener branch in your own fork (I don't understand why does GH even permit that..)


afranzi:
Hi @erosb, this looks awesome!

Do you think should we add some documentation about this new feature in the `Readme.md`? So other developers can start using it?

Thanks for helping me out with this :)

erosb:
Thanks @afranzi I can add the docs (eg. to the readme) after merging / before releasing.

PR #254, opened 2019-01-04
Publish sources jar and javadoc jar

jklukas:
Currently, IntelliJ and other IDEs aren't able to automatically jump to source or documentation for this library, since those jars aren't published.

We should get a nice developer experience improvement by providing these jars. This PR follows the example plugin configuration in https://github.com/jitpack/maven-simple


erosb:
Great, thanks!

Issue #255, opened 2019-01-07
schemaReferenced in ValidationListener not returning the proper location


afranzi:
It seems that from version 1.10.0 to the hash commit `5722c0f4a1306705a56a4d9c11c82bb5709cc57a` some behaviour about how the SchemaReferenced works changed.

In the 1.10.0 the SchemaReferenced was returning the location of the schema referenced:
```
override def schemaReferenced(event: SchemaReferencedEvent): Unit = {
    event.getReferredSchema.getSchemaLocation == "#/allOf/0/properties/product"
}
```

Now, it's returning the schema Id 
```
    event.getReferredSchema.getSchemaLocation == "/schemas/objects/Product/1.json"
```

I've been trying to figure out where was the change, but I didn't manage to found it.

The event, I'm validating is
```
{
  "user": {
    "id": "5a34008f8cece4000764cd5a"
  },
  "device": {
    "id": "5a34008f8cece4000764cd5a",
    "platform": "Android"
  },
  "product": {
    "id": "remixprototype",
    "version": "0.0.1"
  },
  "deploymentEnv": "prod",
  "createdAt":"2018-07-19T13:37:28+02:00",
  "schema": "/schemas/events/device-sensor-event/1.json",
  "source": "sensor",
  "data": {
    "type": "light",
    "level": 0.8
  }
}
```
And the schema is 
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "/schemas/events/base-event/1.json",
  "additionalProperties": true,
  "description": "Base schema for all user-generated events (on device)",
  "properties": {
    "user": {
      "description": "User information",
      "$ref": "/schemas/objects/User/1.json"
    },
    "product": {
      "description": "Product information",
      "$ref": "/schemas/objects/Product/1.json"
    },
    "deploymentEnv": {
      "description": "Deployment environment in use",
      "enum": [
        "dev",
        "test",
        "stage",
        "prod"
      ]
    },
    "createdAt": {
      "description": "Timestamp when the event was generate (following rfc 3339 format)",
      "type": "string",
      "format": "date-time"
    },
    "schema": {
      "description": "Name of the schema to validate against",
      "type": "string"
    },
    "source": {
        "description": "Source of the data point",
        "type": "string",
        "enum": ["sensor", "questionnaire"]
    }
  },
  "required": [
    "source",
    "schema",
    "user",
    "product",
    "deploymentEnv",
    "createdAt"
  ],
  "type": "object"
}
```
[...]

PR #256, opened 2019-01-09
Add Schema validation path traceability

afranzi:
This PR aims to show a PoC of tracing how the schema is being validated, that means which subschemas are being validated regarding the location of them.

The example from https://gist.github.com/afranzi/76497177ea8bc305c355b49bc3b2f6b0 shows how we are using a complex schema to validate one SensorEvent. After validating we generate the following output:
```
// Output after validating the 1.device-sensor-wifi-event.json
SchemaReferenced("#/device", "/schemas/objects/Device/1.json")
SchemaReferenced("#/product","/schemas/objects/Product/1.json")
SchemaReferenced("#/user","/schemas/objects/User/1.json")
SchemaReferenced("#","/schemas/events/base-event/1.json")
SchemaReferenced("#","/schemas/events/base-device-event/1.json")
SchemaReferenced("#/data/scan/[0]","/schemas/objects/sensors/WifiConnection/1.json")
SchemaReferenced("#/data/scan/[1]","/schemas/objects/sensors/WifiConnection/1.json")
SchemaReferenced("#/data","/schemas/objects/sensors/SensorWifi/1.json")
```

Then we know that this event was formed by a SensorWifi data inside the `#/data` field and then inside that one the array `#/data/scan` contains two elements that validates agains the WifiConnection schema. Knowing which schema validates each item matters if the array of items can contain different types of subschemas.

I'm not really proud of this solution by passing all the paths through the `accept` and `validate` methods. So I would be open to any advice, proposal, idea, best practise to improve it.

What do you think @erosb? You can see in the [Gist](https://gist.github.com/afranzi/76497177ea8bc305c355b49bc3b2f6b0) how are we using the `ValidationListener`.

Thanks,
Regards,

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -21,7 +21,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.everit.json</groupId>
     <artifactId>org.everit.json.schema</artifactId>
-    <version>0.0.0-develop</version>
+    <version>2.0.0</version>
     <packaging>bundle</packaging>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/core/src/main/java/org/everit/json/schema/ArraySchema.java
+++ b/core/src/main/java/org/everit/json/schema/ArraySchema.java
@@ -236,8 +236,8 @@ public class ArraySchema extends Schema {
         }
     }
 
-    @Override void accept(Visitor visitor) {
-        visitor.visitArraySchema(this);
+    @Override void accept(Visitor visitor, List<String> path) {
+        visitor.visitArraySchema(this, path);
     }
 
     @Override
--- a/core/src/main/java/org/everit/json/schema/ArraySchemaValidatingVisitor.java
+++ b/core/src/main/java/org/everit/json/schema/ArraySchemaValidatingVisitor.java
@@ -9,7 +9,9 @@ import java.util.List;
 import java.util.Optional;
 import java.util.function.IntFunction;
 import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
+import java.util.List;
 import org.json.JSONArray;
 
 class ArraySchemaValidatingVisitor extends Visitor {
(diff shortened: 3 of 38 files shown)


erosb:
Hello @afranzi how is this newly introduced `path` different from the existing `Schema#getSchemaLocation()` (which is computed during schema loading and also available in visitor methods)?

afranzi:
Hi @erosb,

Using the `Schema#getSchemaLocation()` it returns the following from the `SchemaReferenced` Event

```
SchemaReferenced(/schemas/objects/sensors/WifiConnection/1.json,/schemas/objects/sensors/WifiConnection/1.json)
SchemaReferenced(/schemas/objects/sensors/SensorWifi/1.json,/schemas/objects/sensors/SensorWifi/1.json)
SchemaReferenced(/schemas/objects/Device/1.json,/schemas/objects/Device/1.json)
SchemaReferenced(/schemas/objects/Product/1.json,/schemas/objects/Product/1.json)
SchemaReferenced(/schemas/objects/User/1.json,/schemas/objects/User/1.json)
SchemaReferenced(/schemas/events/base-event/1.json,/schemas/events/base-event/1.json)
SchemaReferenced(/schemas/events/base-device-event/1.json,/schemas/events/base-device-event/1.json)
```

As you can see, it returns the schema location, but not the exact path in the event validated where the subschema was matching.

Regarding your comment in the issue https://github.com/everit-org/json-schema/issues/255#issuecomment-453810737, which proposal or workaround would you propose to get the correct path?

erosb:
Hello @afranzi ,

thinking through it, regarding both your open issues, I think the best I can advise you is keeping up a permanent fork of the library (at least for the duration of your project), altering it in the way you need to. Then when you are done with your project we will see what added features can we backport into this repository.

From my view the situation is that you are using the library for something that it isn't really designed for, and in a way that I don't clearly understand. No doubt it sounds like an awesome research project, but I simply don't have the mental capacity these times to understand it, hence it is hard to advise on design in a forward-thinking way.

Thank you for your understanding.

PR #261, opened 2019-01-14
Report ReferenceSchema Event only when it match the schema

afranzi:
This PR aims to report the ReferecenceSchema event only when the schema is matching against the event.


erosb:
Hi! This change can break existing behaviour (so it can break other users' visitors), can't it?

afranzi:
Yes, this change can break existing behaviour in case that someone was using the SchemaReference Event.

The issue, is that json-schema is reporting the schema referenced even if the schema is not maching. Should we move to use the same approach of Match/UnMatch event with the Referent event?

So we can inform in the Listener if the Referenced Schema is matching or not?

Issue #259, opened 2019-01-13
unprocessedProperties not populating for Fields referencing another definition


NaveenGurram:
I was looking to leverege recenlty added "unprocessedProperties" but I see it is not loading for a field when it is referencing custom definition.

When I load below schema I was expecting  "dateOfBirth" RefSchema has unprocessedProperties value of "unproc7". 

is it possible to load unprocessedProperties for a field in properties when referencing another definition? 

**JSON Schema**
```json
{
    "title": "ttt",
    "description": "descr",
    "$id": "http://example.org",
    "default": 22,
    "nullable": false,
    "readOnly": false,
    "writeOnly": false,
    "definitions": {
      "monthDayYearDate": {
        "$comment": "Represents the MM/DD/YYYY date format.",
        "description": "(MM/dd/yyyy)",
        "type": "string",
        "maxLength": 10,
        "pattern": "^\\d{2}/\\d{2}/\\d{4}$"
      }
    },
    "type": "object",
    "unproc0": 1,
    "unproc1": "asdasd",
    "properties": {
      "prop": {
        "type": "string"
      },
      "prop2": {
        "unproc4": true,
        "unproc5": null
      },
      "prop3": {
        "$ref": "#",
        "unproc6": false
      },
      "dateOfBirth": {
        "title": "Date of Birth",
        "$ref": "#/definitions/monthDayYearDate",
        "unproc7": {
          "abc": "message 1"
        }
      }
    }
```

erosb:
Good question :)
I think in both references the unprocessed properties should be present in the `ReferenceSchema` instance but not in the referred schema. I didn't try it out (though while reading `SchemaLoader#loadSchemaObject()` and `SchemaLoader#runSchemaExtractors()` I can't see where the control flow would be different).

NaveenGurram:
@erosb  if you think it's the right behavior I can look at the code and issue a PR if I can solve the problem.

erosb:
@NaveenGurram please proceed (and include at least one test in the PR), thank you.

NaveenGurram:
@erosb  Thanks for providing the fix, but the fix is not working especially when there are more elements referring to same ref element.  In the below example for "prop5" I was expecting "unproc09" in unprocessed properties but it is missing and addition to that it is showing the property from "prop4" in prop5. 

```  "schemaRefWithUnprocessedProperties": {
    "title": "ttt",
    "description": "descr",
    "$id": "http://example.org",
    "default": 22,
[...]

PR #266, opened 2019-01-23
[259] fix to support populating unprocessedProperties for elements ha…

NaveenGurram:
https://github.com/everit-org/json-schema/issues/259 
SchemaLoader is loading properly, but somewhere when setting it is missing out unprocessedProperties.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/ReferenceSchema.java
+++ b/core/src/main/java/org/everit/json/schema/ReferenceSchema.java
@@ -34,6 +34,8 @@ public class ReferenceSchema extends Schema {
             if (retval == null) {
                 retval = new ReferenceSchema(this);
             }
+            // adding unprocessed properties from the parent to the actual referenced schema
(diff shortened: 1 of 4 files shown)


NaveenGurram:
@erosb would be able to check out the fix to verify if this is the right manner of fixing this issue?

erosb:
Hello @NaveenGurram , thank you for raising the PR.
In the end I decided to fix it in a different way, now `ReferenceSchema` overrides the unprocessedProperties handling of `Schema`. This is not a good solution but at least it keeps mutability more local (instead of making `Schema` mutable also with the `appendUnprocessedProperties()` method ).

Thanks for your time spent on investigating this, the tests were useful, even if the final fix was different.

NaveenGurram:
@erosb  i fixed it locally and it is not very elegant, so I am trying to optimize the fix to fit your existing visibility model and make sure i test out thoroughly, in couple of days, I will refine the code and will issue a PR.

PR #274, opened 2019-02-05
Format support for all Objects

srivastavr:
Format support for all Objects(Integer,Number,Boolean) ,rather than only String


erosb:
This branch doesn't compile. Closing the PR.

PR #293, opened 2019-04-11
fix application of Apache 2.0 license

MarkAtwood:
correct the adhoc LICENSE.md file to a correct pair of LICENSE and NOTICE files.

This does not change the license on the project.
It does make it easier for automatic code license analyzer tools.


erosb:
Thank you @MarkAtwood , this is a useful improvement. Also sorry about the late response.

Issue #306, opened 2019-05-30
HTTPS $schema endpoint causes SchemaException


rjlohan:
Although the draft JSON Schema documents are available at both `https://json-schema.org` and `http://json-schema.org`, SchemaLoader throws `SchemaException` if the HTTPS endpoint is used; https://github.com/everit-org/json-schema/blob/506205044e334793724e89502c359407faa62dbd/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java#L311

PR #307, opened 2019-05-30
[Issue #306] Supports HTTPS endpoint for version inference

rjlohan:
Fixes Issue #306 

Schemas hosted on json-schema.org can be accessed at both HTTP and HTTPS endpoints. The current inference logic for schema versions caused SchemaException if HTTPS endpoint was used in the $schema


erosb:
It may be more concise to perform this check in a protocol-relative way, but this fix is good-enough for now, and it doesn't introduce any API change that might be difficult to maintain in the future, so LGTM.
Thanks for your contribution!

PR #315, opened 2019-07-12
 schema and subject

himkota:
Hi 
I have created a pull request for the issue 314.
The causing exceptions currently are coming with two schema validation failures expectation is to get only one.


erosb:
Hello,

thank you for the bugreport and PR. This is caused by the way how unrelated keywords (`"properties"` and conditionals, in this case) are modeled by the library (they are wrapped with an implicit `"allOf"` during loading) and this can't be fixed without breaking backward compatibility.

PR #332, opened 2019-10-09
error handling for classpath schema client

rjruizes:
Currently, if you reference a schema file that does not exist, you get this error message:
```
unknown protocol: classpath
```
because the `classpath` URL is being passed to the fallback client, which can't process it.

This PR improves the error message to name the classpath resource that cannot be found:
```
Caused by: java.io.IOException: Could not find classpath://com/test/predefined.json
```


erosb:
Nice fix, thank you for your work @rjruizes !

PR #333, opened 2019-10-11
Allow setting root document uri in SchemaLoader

tmarsteel:
I need an absolute path to all schemas i load; Setting it afterwards doesn't cut it because it won't apply to nested schemas.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
+++ b/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java
@@ -265,6 +265,11 @@ public class SchemaLoader {
             enableOverrideOfBuiltInFormatValidators = true;
             return this;
         }
+
+        public SchemaLoaderBuilder rootDocumentLocation(SchemaLocation location)  {
+            pointerToCurrentObj(location);
+            return this;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


tmarsteel:
Hi :)

I use this code to load the schemas:

```kotlin
val loader = SchemaLoader.builder()
                        .resolutionScope("file:${schemaPath.parent}/")
                        .schemaJson(schemaJson)
                        .build()
val schema = loader.load().build()
```

With that, `schema.getLocation()` does have `rootDocumentURI == null`. Did i miss something?

erosb:
Hello, this problem makes sense, though probably we need an other approach to fix it. So currently we already have 2 ways for setting the location of a schema: `SchemaLoader#resolutionScope()` and `SchemaLoader#pointerToCurrentObj()` which is used internally by the library. I don't want to add a third builder method which does something similar. I think the best solution would be changing `SchemaLoader#resolutionScope(URI)` to set both `this.id` and `this.pointerToCurrentObj` in the builder method. Do you want to give it a go?

erosb:
Nevermind, I fixed it in 993c3f7798de21f94c403d391a20dca7e9e325af
Closing this PR. Please let me know if you have any further questions.

PR #356, opened 2020-03-16
provide examples field to set example value for a field.

sreekesh93:
Add feature to set examples for each json field as part of schema

### DIFF
--- a/core/src/main/java/org/everit/json/schema/Schema.java
+++ b/core/src/main/java/org/everit/json/schema/Schema.java
@@ -1,104 +1,22 @@
 package org.everit.json.schema;
 
-import static java.util.Collections.unmodifiableMap;
+import org.everit.json.schema.internal.JSONPrinter;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/everit/json/schema/Schema.java
```diff
@@ -1,104 +1,22 @@
...
-
-        public Builder<S> unprocessedProperties(Map<String, Object> unprocessedProperties) {
-            this.unprocessedProperties = unprocessedProperties;
-            return this;
-        }
-
-        public abstract S build();
-
-    }
+    private final String examples;
```
erosb: A parsed `Object` would be much more useful than a string.

sakthivel-tw:
@erosb Thank you for the awesome library. Regarding the `examples` keyword, the spec says it was introduced in Draft 6 ?

https://json-schema.org/understanding-json-schema/reference/generic.html?highlight=examples

PR #366, opened 2020-04-25
Add ProviderValidators to Schema Builder for customize the behavior of CustomValidator dynamically

GiskardB:


erosb:
Hello @GiskardB , is this change related to any parts of the json schema specfication?

GiskardB:
Hi @erosb,  no it's a simple feature of the this library that add custom validator in runtime.

CrCat:
Nice! This is exactly what I'm looking for to support the flexibility of customized format validation!

PR #367, opened 2020-05-02
Consider two numbers equal if they have the same integer part and dif…

zgyorffi:
…ers only in the Number type and the number of zeros as decimals.

As org.json library parses a number as the following types the comparison works fine for these:
* BigDecimal
* BigInteger
* Long
* Integer
* Double

In case of floats scientific notation is used in String representation for values >= 10^7 so the comparison fails, for doubles it is handled.

This PR addresses the issue which causes 6 test cases to fail in the draft 2019 test set. E.g. 1 should be equal to 1.0.


core/src/main/java/org/everit/json/schema/ObjectComparator.java
```diff
@@ -1,20 +1,32 @@
...
  * Deep-equals implementation on primitive wrappers, {@link JSONObject} and {@link JSONArray}.
  */
 public final class ObjectComparator {
 
+    private static NumberFormat numberFormat = NumberFormat.getInstance();
+
+    static {
+        numberFormat.setMaximumFractionDigits(Integer.MAX_VALUE);
+        numberFormat.setGroupingUsed(false);
+    }
```
erosb, line 25: Format classes are not thread safe. Please let's create a separate instance whenever it is used (so a `private static NumberFormat createNumberFormat()` method).

core/src/main/java/org/everit/json/schema/ObjectComparator.java
```diff
@@ -33,6 +45,15 @@ public static boolean deepEquals(Object obj1, Object obj2) {
                 return false;
             }
             return deepEqualObjects((JSONObject) obj1, (JSONObject) obj2);
+        } else if (obj1 instanceof Number) {
+            if (!(obj2 instanceof Number)) {
+                return false;
+            } else if (!obj1.getClass().equals(obj2.getClass())) {
+                String obj1AsString = obj1 instanceof Double && (Double) obj1 >= 10000000 ? numberFormat.format(obj1) : JSONObject.numberToString((Number) obj1);
```
erosb, line 52: Why did you choose `10000000` as the treshold? Is there any specific reason?

core/src/main/java/org/everit/json/schema/ObjectComparator.java
```diff
@@ -33,6 +45,15 @@ public static boolean deepEquals(Object obj1, Object obj2) {
...
+                String obj2AsString = obj2 instanceof Double && (Double) obj2 >= 10000000 ? numberFormat.format(obj2) : JSONObject.numberToString((Number) obj2);
```
erosb, line 53: If the number is < 10000000 then it gets cast twice (once to Double, then to Number). Can we please avoid the duplicate casting? It will need an explicit if block.

Also the code is duplicated for `obj1` and `obj2` so please extract it to a `numberToString()` method or something similar.

core/src/main/java/org/everit/json/schema/ObjectComparator.java
```diff
@@ -33,6 +45,15 @@ public static boolean deepEquals(Object obj1, Object obj2) {
                 return false;
             }
             return deepEqualObjects((JSONObject) obj1, (JSONObject) obj2);
+        } else if (obj1 instanceof Number) {
+            if (!(obj2 instanceof Number)) {
+                return false;
+            } else if (!obj1.getClass().equals(obj2.getClass())) {
```
erosb, line 51: Plese compare class instances with `!=`

erosb, line 51: Please compare `Class` instances with `!=`

core/src/main/java/org/everit/json/schema/ObjectComparator.java
```diff
@@ -33,6 +45,15 @@ public static boolean deepEquals(Object obj1, Object obj2) {
...
+                String obj1AsString = obj1 instanceof Double && (Double) obj1 >= 10000000 ? numberFormat.format(obj1) : JSONObject.numberToString((Number) obj1);
+                String obj2AsString = obj2 instanceof Double && (Double) obj2 >= 10000000 ? numberFormat.format(obj2) : JSONObject.numberToString((Number) obj2);
+
```
erosb, line 54: If the number is < 10000000 , then it gets cast twice (once to `Double`, then to `Number`). Please let's cast only once. Also please extract this string conversion to a separate method, because it is duplicated for `obj1` and `obj2`.

Why did you choose `10000000` as the treshold? Is there any specific reason?

zgyorffi: I chose 10000000 because the numbers larger than or equal to it are represented in scientific notation but numbers smaller than 0.001 are also represented in scientific notation those should be handled as well. Anyway I decided to implement a different approach because NumberFormat is determined by Locale. I could bind the format to a specific Locale but still there would be a small chance that the formatting behaves in an unexpected way in some special cases that I do not see now.

PR #374, opened 2020-07-05
Skip validation of elements in array against additionalItems schema i…

zgyorffi:
…f "items" schema is missing

### DIFF
--- a/core/src/test/java/org/everit/json/schema/ArraySchemaTest.java
+++ b/core/src/test/java/org/everit/json/schema/ArraySchemaTest.java
@@ -42,6 +42,13 @@ public class ArraySchemaTest {
     }
 
     @Test
+    public void ignoreAdditionalItemsIfNoItemsSchema() {
+        ArraySchema.builder()
+            .schemaOfAdditionalItems(FalseSchema.INSTANCE)
+            .build().validate(ARRAYS.get("noItemSchema"));
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/everit/json/schema/ArraySchemaTest.java
```diff
@@ -41,6 +41,13 @@ public void additionalItemsSchema() {
                 .build().validate(ARRAYS.get("additionalItemsSchema"));
     }
 
+    @Test
+    public void ignoreAdditionalItemsIfNoItemsSchema() {
+        ArraySchema.builder()
+            .schemaOfAdditionalItems(FalseSchema.INSTANCE)
+            .build().validate(ARRAYS.get("noItemSchema"));
```
erosb, line 48: Hello @zgyorffi is the "noItemSchema" present in the json file? I can't see it in the commit.

zgyorffi: Hi @erosb!
I reused an existing one:
https://github.com/everit-org/json-schema/blob/a8dc6860bd6a918d729cceac7bc46b2c7c871cfc/core/src/test/resources/org/everit/jsonvalidator/arraytestcases.json#L2

If you prefer it I can create a separate one.

erosb: thanks, I grepped in the wrong directory...

Issue #302, opened 2019-05-22
StackOverflowError when calling hashCode on mutually recursive schemas


cpchen:

tmarsteel:
Would it suffice to have `ReferenceSchema` not include the referred schema into the hashCode? Actually, that would make a lot of sense to me. The referred schema depends on the schema loader pretty much and the correctness of `ReferenceSchema`s equals+hashCode shouldn't depend on the schema loader.

erosb:
Hello @tmarsteel , you can give it a go, but in that case I'm afraid the `equals()`will still run into an infinite recursion, so not sure if that helps a lot. Feel free to raise a PR with the fix and a passing test, if you want. Thanks for your idea anyway.

rayokota:
We use this library in Confluent Schema Registry, and are seeing this error as well when `equals` or `hashCode` is called on a recursive schema.  https://github.com/confluentinc/schema-registry/issues/1549

shionefaye:
@rayokota  [Conflunece Schema Registry](https://github.com/confluentinc/schema-registry/issues/1549) issue on recursive schema has been fixed and merged recently. What is needed to be done in this package?

rayokota:
@shionefaye , the issue still exists in this library.  For Schema Registry the code has been changed to compare instances of `JsonNode` from Jackson rather than instances of `Schema` from this library.

shionefaye:
> @shionefaye , the issue still exists in this library. For Schema Registry the code has been changed to compare instances of `JsonNode` from Jackson rather than instances of `Schema` from this library.

Got you. I misread your early message and thought this package depends on Conflunece Schema Registry.

PR #378, opened 2020-07-17
#302: do not use ReferenceSchema#referredSchema for equals+hashCode

tmarsteel:
This solves Issue #302. Not including the referred schema feels right to me since it depends on the schemaLoader. I have no idea whether people rely on referredSchema being in equals+hashCode, though.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/ReferenceSchema.java
+++ b/core/src/main/java/org/everit/json/schema/ReferenceSchema.java
@@ -145,7 +145,6 @@ public class ReferenceSchema extends Schema {
             return that.canEqual(this) &&
                     Objects.equals(refValue, that.refValue) &&
                     Objects.equals(unprocessedProperties, that.unprocessedProperties) &&
-                    Objects.equals(referredSchema, that.referredSchema) &&
                     Objects.equals(title, that.title) &&
                     Objects.equals(description, that.description) &&
                     super.equals(that);
@@ -156,7 +155,7 @@ public class ReferenceSchema extends Schema {
 
     @Override
     public int hashCode() {
-        return Objects.hash(super.hashCode(), referredSchema, refValue, unprocessedProperties, title, description);
+        return Objects.hash(super.hashCode(), refValue, unprocessedProperties, title, description);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/everit/json/schema/ReferenceSchema.java
```diff
@@ -145,7 +145,6 @@ public boolean equals(Object o) {
             return that.canEqual(this) &&
                     Objects.equals(refValue, that.refValue) &&
                     Objects.equals(unprocessedProperties, that.unprocessedProperties) &&
-                    Objects.equals(referredSchema, that.referredSchema) &&
```
erosb, line 148: Hello @tmarsteel , do you think we loose anything if we keep this line? So my concern is that, with this change, there can be (rare) cases when two schemas are considered equal when they are not (I mean eg. two references pointing to `#` , but this denotes the roots of two different schema documents). 
What do you think about keeping this line? I see that it might run into infinite `equals()` recursions, but that's probably a very rare case.

rayokota: In the case of Confluent Schema Registry, we call `equals` quite often on schema.   So we would prefer the fix above.

tmarsteel: That's a very good point. The testcase and my actual cyclic schemas both work with `referredSchema` being included in equals.

tmarsteel:
Hm, this doesn't suffice yet. The `HashCodeRecursionTest` succeeds in my IDE but fails in maven. Looking into it.

tmarsteel:
Okay, so it turns out that equals will always cause a stackoverflow when you call it on equal (but not identical) schemas. I have such a cycle in one of my projects, so keeping `referredSchema` in equals doesn't work for me. I see two solutions for me here :D
1. make sure every schema file i have is not loaded more than once so the `equals` method can take the identity-shortcut
2. Since `ReferenceSchema` is the only way to create cycles in json schemas, these cycles can be detected and handled there. I have drafted something (13438c94) that passes the tests, but i'm very unsure whether this code should go productive anywhere

rayokota:
@tmarsteel @erosb We would prefer the fix where `referredSchema` is excluded from both `equals` and `hashCode`.  

If a client wants to additionally compare the `referredSchema`, the client can always do that check "manually".  Optionally perhaps a `deepEquals` method can be provided that does that.    However, without this PR, there is no way to avoid the `StackOverFlowError`.

In other words, we need one method to avoid the `StackOverFlowError`.  Optionally, we can provide a separate method `deepEquals`, which might still result in the error.

tmarsteel:
@rayokota that sounds very reasonable to me. I'm okay with implementing it. It would be a breaking change, so likely the release will not be that soon, will it? @erosb

PR #384, opened 2020-09-15
Support dependentRequired instead of dependencies

zgyorffi:
Replaces dependencies keyword to dependentRequired as suggested by draft-2019-09

### DIFF
--- a/core/src/main/java/org/everit/json/schema/loader/ObjectSchemaLoader.java
+++ b/core/src/main/java/org/everit/json/schema/loader/ObjectSchemaLoader.java
@@ -1,11 +1,13 @@
 package org.everit.json.schema.loader;
 
 import static java.util.Objects.requireNonNull;
+import static org.everit.json.schema.loader.SpecificationVersion.DRAFT_201909;
 import static org.everit.json.schema.loader.SpecificationVersion.DRAFT_6;
 
 import org.everit.json.schema.ObjectSchema;
 import org.everit.json.schema.Schema;
 import org.everit.json.schema.regexp.Regexp;
+
 
 /**
  * @author erosb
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/everit/json/schema/loader/ObjectSchemaLoaderTest.java
```diff
@@ -56,12 +56,16 @@ public void objectWithAdditionalPropSchema() {
     public void objectWithPropDep() {
         ObjectSchema actual = (ObjectSchema) SchemaLoader.load(get("objectWithPropDep"));
         assertEquals(1, actual.getPropertyDependencies().get("isIndividual").size());
+        actual = (ObjectSchema) loadAsV201909(get("objectWithPropDep_V201909"));
```
erosb, line 59: This can be movd into a separate tests, so that each test contains only one given-when-then triad (same for the other test changes below).

core/src/test/java/org/everit/json/schema/loader/DefinesPropertyTest.java
```diff
@@ -113,7 +114,7 @@ public void testOfTest() {
 
     @Test
     public void patternPropsAndSchemaDefs() {
-        ObjectSchema actual = (ObjectSchema) SchemaLoader.load(get("patternPropsAndSchemaDeps"));
+        ObjectSchema actual = (ObjectSchema) loadAsV201909(get("patternPropsAndSchemaDeps"));
```
erosb, line 117: Please try to avoid changing existing tests, add a new one instead (same comment for the above tests in the class and in `testschemas.json`).

core/src/main/java/org/everit/json/schema/loader/ObjectSchemaLoader.java
```diff
@@ -29,38 +31,43 @@ public ObjectSchemaLoader(LoadingState ls, LoaderConfig config, SchemaLoader def
...
                 });
-        ls.schemaJson().maybe("dependencies").map(JsonValue::requireObject)
+            });
+        if (ls.specVersion().isAtLeast(DRAFT_201909)) {
+            ls.schemaJson().maybe("dependentRequired").map(JsonValue::requireObject)
+                .ifPresent(deps -> addDependencies(builder, deps));
+        } else {
+            ls.schemaJson().maybe("dependencies").map(JsonValue::requireObject)
                 .ifPresent(deps -> addDependencies(builder, deps));
+        }
```
erosb, line 56: Instead of handling it here, I think it would be more consistent to add a `String dependentRequiredKeyword()` abstract method to the `SpecificationVersion` enum, which returns `"dependentRequired"` for draft201909,  and `"required"` for all the older enum values. I suggest this because `SpecificationVersion#idKeyword()` serves a similar purpose, so it would be a consistent approach.

erosb (changes requested):
Thanks for raising the PR @zgyorffi , good direction overall. I added a couple of comments.

core/src/main/java/org/everit/json/schema/loader/ObjectSchemaLoader.java
zgyorffi: Cool, I did not notice `SpecificationVersion#idKeyword()`

core/src/test/java/org/everit/json/schema/loader/DefinesPropertyTest.java
zgyorffi: My understanding was that the purpose of these tests to test `definesProperty` method and the emphasis is not on the loading which is tested separately in ObjectSchemaLoaderTest.java. That is why I finally did not duplicate the tests just thought to update the loading to the latest specification.

core/src/test/java/org/everit/json/schema/loader/ObjectSchemaLoaderTest.java
zgyorffi: Ok, at first I used separate tests then I changed my mind :)

erosb (approved):
This looks great now, thank you for your contribution @zgyorffi !

Issue #381, opened 2020-07-22
expected type: Number, found: String, even though schema requires an Integer


bilalba:
For the following schema
```
{
  "$id": "https://example.com/person.schema.json",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "integer",
      "description": "The person's first name."
    }
  }
}
```

and this object:  `{"firstName":"1.123"}`

The error that I get is:

> expected type: Number, found: String

when it should really say:

> expected type: Integer, found: String

I'm working on type-casting the object values to the correct type defined in the schema, and when I see the above error, it's ambiguous whether I should type-cast it to an integer or a number.

erosb:
Hello, this is now fixed by @zgyorffi in #387 . Thanks for reporting the issue.

bilalba:
Thanks @zgyorffi @erosb

PR #387, opened 2020-10-18
Print Integer as expected type if String is given

zgyorffi:
Addressing #381

### DIFF
--- a/core/src/main/java/org/everit/json/schema/NumberSchema.java
+++ b/core/src/main/java/org/everit/json/schema/NumberSchema.java
@@ -150,7 +150,7 @@ public class NumberSchema extends Schema {
         return requiresInteger;
     }
 
-    public boolean requiresNumber() {
+    public boolean isRequiresNumber() {
         return requiresNumber;
     }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/everit/json/schema/NumberSchema.java
```diff
@@ -150,7 +150,7 @@ public boolean requiresInteger() {
         return requiresInteger;
     }
 
-    public boolean isRequiresNumber() {
+    public boolean requiresNumber() {
```
erosb, line 153: Please don't rename public methods, it breaks API compatibility (I know that `isRequiresNumber()` is a very poor name :))

core/src/main/java/org/everit/json/schema/TypeChecker.java
```diff
@@ -0,0 +1,21 @@
+package org.everit.json.schema;
+
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+
+public class TypeChecker  {
```
erosb, line 10: Please make this class package-private.

erosb (commented):
Minor comments added, otherwise looks good, thank you @zgyorffi !

PR #390, opened 2020-11-17
schema builder fluency (fixes #303)

### LINKED ISSUE #303: Base Schema builder is returned in some builder methods.
When trying to build a concrete schema using it's builder I met a strange situation when method calling orders matters because of their return types.

For example this code is valid:

```
        StringSchema stringValue = StringSchema.builder()
            .maxLength(10)
            .title("String value")
            .build();
```

while this is not:

```
        StringSchema stringValue = StringSchema.builder()
            .title("String value")
            .maxLength(10)
            .build();
```

This feels like a bug to me.

One of the possible solutions is to use a concrete builder as a generic parameter:

```
    public abstract class Schema {
        public abstract static class Builder<S extends Schema, B extends Builder> {
            public B title(String title) {
                this.title = title;
                //noinspection unchecked
                return (B) this;
            }
        }
    }

    public class StringSchema extends Schema {
        public static class Builder extends Schema.Builder<StringSchema, Builder> {

        }
    }
```

JoshMcCullough:
**Breaking Changes**

`Schema.Builder` needs to return the subclassed builder instance rather than `Builder<S>` in order to allow for fluency while building schemas. See related issue.

Note: I also renamed `ConstSchema.ConstSchemaBuilder` to just `Builder` to match the other builders for consistency. I can pull this change back out if you don't want it. Since this PR is already a breaking change, I figured it made sense.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -167,6 +167,12 @@
                                 <sourceCompatible>true</sourceCompatible>
                             </overrideCompatibilityChangeParameter>
                             <overrideCompatibilityChangeParameter>
+                                <compatibilityChange>METHOD_REMOVED</compatibilityChange>
+                                <binaryCompatible>true</binaryCompatible>
+                                <sourceCompatible>true</sourceCompatible>
+                                <semanticVersionLevel>MAJOR</semanticVersionLevel>
+                            </overrideCompatibilityChangeParameter>
+                            <overrideCompatibilityChangeParameter>
                                 <compatibilityChange>METHOD_RETURN_TYPE_CHANGED</compatibilityChange>
                                 <binaryCompatible>true</binaryCompatible>
                                 <sourceCompatible>true</sourceCompatible>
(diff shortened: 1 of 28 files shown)


JoshMcCullough:
I'm trying to figure out how to instruct japicmp that it's okay that I renamed that class since this would be released as a new major...

JoshMcCullough:
Okay, updated the japicmp config to allow these changes for a major bump.

erosb:
Hello @JoshMcCullough , thank you for your work on this. I'm afraid we won't break backward-compatibility and increment the major version number for the purpose of this single fix (and other potential breaking changes are not under discussion right now).

JoshMcCullough:
@erosb I understand the hesitance. And I can work with it using my own fork. I will just point out that creating a fluent schema as it is now is _possible_. But a general user would be confused if they tried it. I hope you'll bring this in at some point. I don't really see the hesitation in releasing a new major. They are just numbers, after all, and that's the point of semantic versioning. I would understand if you wanted to wait before doing so, but hope it's on your roadmap.

erosb:
Yes, I thought about the same fix years ago. Thanks for the PR. Let's leave it open for now, and we will merge it when 2.0 is on the line.

JoshMcCullough:
Opened new PR: https://github.com/everit-org/json-schema/pull/465

PR #401, opened 2021-01-26
Fix #400, keep precision in case of big numbers

### LINKED ISSUE #400: maximum field does not work properly with numbers around 1.0E16
schema:
`
{   
    "type": "number",
    "maximum": 9999999999999999    
}
`
payload:
`10000000000000000` or `10000000000000001`
This does not return a maximum value exceeded validation. 

**Note:** When the payload is `10000000000000002` the validation is returned correctly.

**Observations**
From debugging the code i found that when calling .doubleValue() in https://github.com/everit-org/json-schema/blob/master/core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java#L67 the number is truncated to 1.0E16 as well as the numberSubject value.

zgyorffi:


core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java
```diff
@@ -41,17 +42,17 @@ void visitMinimum(Number minimum) {
         if (minimum == null) {
             return;
         }
-        if (exclusiveMinimum && numberSubject <= minimum.doubleValue()) {
+        if (exclusiveMinimum && compare(numberSubject, minimum) <= 0) {
             owner.failure(subject + " is not greater than " + minimum, "exclusiveMinimum");
-        } else if (numberSubject < minimum.doubleValue()) {
+        } else if (compare(numberSubject, minimum) < 0) {
```
erosb, line 47: Can you please extract the `compare(numberSubject, minimum)` value into a local variable?

core/src/main/java/org/everit/json/schema/NumberSchemaValidatingVisitor.java
```diff
@@ -62,9 +63,9 @@ void visitMaximum(Number maximum) {
         if (maximum == null) {
             return;
         }
-        if (exclusiveMaximum && maximum.doubleValue() <= numberSubject) {
+        if (exclusiveMaximum && compare(maximum, numberSubject) <= 0) {
             owner.failure(subject + " is not less than " + maximum, "exclusiveMaximum");
-        } else if (maximum.doubleValue() < numberSubject) {
+        } else if (compare(maximum, numberSubject) < 0) {
```
erosb, line 68: Can you please extract the `compare(maximum, numberSubject)` value into a local variable?

erosb (changes requested):
Thank you for the thorough test coverage @zgyorffi ! I added two minor comments.

PR #405, opened 2021-02-26
Issue 404: Non-deterministic validation result when useDefaults are u…

paulfrmbstn:
…sed with combined schema dependencies

### DIFF
--- a/core/src/main/java/org/everit/json/schema/CombinedSchema.java
+++ b/core/src/main/java/org/everit/json/schema/CombinedSchema.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * Validator for {@code allOf}, {@code oneOf}, {@code anyOf} schemas.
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/everit/json/schema/CombinedSchema.java
```diff
@@ -174,8 +175,20 @@ public ValidationCriterion getCriterion() {
...
+
+        Collection<Schema> combined = subschemas.stream()
+                .filter(schema -> (schema instanceof CombinedSchema))
+                .collect(Collectors.toList());
+
+        Collection<Schema> others = subschemas.stream()
+                .filter(schema -> !(schema instanceof CombinedSchema))
+                .collect(Collectors.toList());
+
+        combined.addAll(others);
```
erosb, line 189: As far as I understand here you add the `others` to a list that was returned by `Collectors.toList()`. The problem with that is, there is no guarantee for the mutability of this list. As [the javadoc says](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toList()):

> There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned;

So please replace this method implementation with any of the followings:
 - a plain old `for` loop and collecting into two separate ArrayLists
 - `Collectors#toCollection(Supplier)` so that you can make sure your stream collects into a mutable list
 -  or you can have only one stream and sort it based on a comparator sorting by an `instanceof CombinedSchema` expression

erosb (changes requested):
Thank you for reporting the issue and raising the PR. It is a bit hacky solution and similar problems can arise later, but if it works for you I'm fine with it. I added a change request.

erosb:
This looks good now, thank you for addressing my comments.

core/src/main/java/org/everit/json/schema/CombinedSchema.java
paulfrmbstn: Updated the PR with requested changes and added more test coverage. Took the comparator approach.

Issue #407, opened 2021-03-19
SchemaException(java.lang.String, java.lang.Exception) called with incorrect argument


SvaponiAkelius:
`org.everit.json.schema.SchemaException#SchemaException(java.lang.String, java.lang.Exception)` is called with first argument as message instead of schemaLocation. Example `new SchemaException("failed to parse " + resp, e);`.

See
- com/github/everit-org/json-schema/org.everit.json.schema/1.12.2/org.everit.json.schema-1.12.2-sources.jar!/org/everit/json/schema/loader/JsonPointerEvaluator.java:93
- com/github/everit-org/json-schema/org.everit.json.schema/1.12.2/org.everit.json.schema-1.12.2-sources.jar!/org/everit/json/schema/loader/internal/JSONPointer.java:86

erosb:
Thanks for reporting it, I created #408 with a fix.

PR #408, opened 2021-03-21
Fixing #407

erosb:
Fixing #407: incorrect SchemaException#schemaLocation when a remote shema fails to parse.

### DIFF
--- a/core/src/test/java/org/everit/json/schema/loader/JsonPointerEvaluatorTest.java
+++ b/core/src/test/java/org/everit/json/schema/loader/JsonPointerEvaluatorTest.java
@@ -32,7 +32,7 @@ public class JsonPointerEvaluatorTest {
     private static final ResourceLoader LOADER = ResourceLoader.DEFAULT;
 
     @Test
-    public void sameDocumentSuccess() {
+    void sameDocumentSuccess() {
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/everit/json/schema/loader/JsonPointerEvaluatorTest.java
```diff
@@ -94,9 +94,21 @@ public void remoteDocumentSuccess() throws URISyntaxException {
...
     }
+    
+    @Test
+    void remoteDocument_jsonParsingFailure() {
+        SchemaClient schemaClient = mock(SchemaClient.class);
+        when(schemaClient.get("http://localhost:1234/hello")).thenReturn(asStream("unparseable"));
+        try {
+            JsonPointerEvaluator.forURL(schemaClient, "http://localhost:1234/hello#/foo", createLoadingState(schemaClient, "")).query();
+            fail("did not throw exception for unparseable json");
+        } catch (SchemaException e) {
```
zgyorffi, line 105: As junit5 has builtin assertThrows method maybe it is better to use that for testing exceptions.

erosb: Makes sense, fixed.

Issue #431, opened 2021-07-30
Support specifying $schema with no dialect


blacktooth:
JSON Schema allows for `$schema` to be `http://json-schema.org/schema#` which defaults to draft4 specification ([Ref](https://json-schema.org/understanding-json-schema/reference/schema.html))

Right now, `SchemaLoader` fails to parse a schema with `$schema` set to `http://json-schema.org/schema#`,

```
org.everit.json.schema.SchemaException: #: could not determine version
at org.everit.json.schema.loader.SchemaLoader.<init>(SchemaLoader.java:338)
...
...
```
**Code**

```
String schema = "{"
            + "\"$schema\": \"http://json-schema.org/schema#\", "
            + "\"type\":\"object\", "
            + "\"properties\":{\"f1\":{\"type\":\"string\", \"maxLength\":10}}"
            + "}";
SchemaLoader.load(new JSONObject(schema));

```

[PR](https://github.com/everit-org/json-schema/pull/432)

PR #432, opened 2021-07-30
Add support for non-dialect specific schema URLs

blacktooth:
This is a fix for issue #431


erosb:
Thank you, nice catch.

PR #443, opened 2021-10-30
ArraySchema#definesProperty()

### LINKED ISSUE #442: SchemaArray's definesProperty implementation
We have a use case of checking whether a property is defined in json schema. My target is to check if the field `type` is defined in the schema, and the field `type` is in an object of an array `adjustments` using [definesProperty](https://github.com/everit-org/json-schema/blob/master/core/src/main/java/org/everit/json/schema/Schema.java#L197-L199) method.

- Here is the Json schema sample: `{
  "type": "object",
  "title": "title",
  "properties": {
    "adjustments": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "default": "hello"
          }
        },
        "required": [
          "type"
        ]
      }
    }
  }
}`

After trying some json pointers (i.e `schema.definesProperty("#/adjustments/0/type") == false`) with no luck, I have looked into the code and found that [ArraySchema](https://github.com/everit-org/json-schema/blob/master/core/src/main/java/org/everit/json/schema/ArraySchema.java) has no implementation but inherit with default false for method `definesProperty`. Any chance we can have this one implemented so that we can check for defined properties in an ArraySchema?

Thank you!

erosb:
Fixes #442

### DIFF
--- a/core/src/main/java/org/everit/json/schema/Schema.java
+++ b/core/src/main/java/org/everit/json/schema/Schema.java
@@ -206,7 +206,7 @@ public abstract class Schema {
      *     <li>{@code definesProperty("#/list/any/prop")} returns {@code true}</li>
      *     <li>{@code definesProperty("#/list/all/prop")} returns {@code true}</li>
      *     <li>{@code definesProperty("#/list/1/prop")} returns {@code true}</li>
-     *     <li>{@code definesProperty("#/list/1/nonexistent")} returns {@code true} (the property is not present in the
+     *     <li>{@code definesProperty("#/list/1/nonexistent")} returns {@code false} (the property is not present in the
      *     subschema)</li>
      *     <li>{@code definesProperty("#/list/8/prop")} returns {@code false} (the {@code "list"} does not define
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/everit/json/schema/Schema.java
```diff
@@ -187,6 +199,18 @@ public void validate(Object subject) {
...
      * </code>
      * </pre>
+     *
+     * You can also check if a subschema of an array defines a property. In that case, to traverse the array, you can either use
+     * an integer array index, or the {@code "all"} or {@code "any"} meta-indexes. For example, in the above schema
+     * <ul>
+     *     <li>{@code definesProperty("#/list/any/prop")} returns {@code true}</li>
+     *     <li>{@code definesProperty("#/list/all/prop")} returns {@code true}</li>
+     *     <li>{@code definesProperty("#/list/1/prop")} returns {@code true}</li>
+     *     <li>{@code definesProperty("#/list/1/nonexistent")} returns {@code true} (the property is not present in the
```
vokhanhhuy, line 209: should return false

PR #445, opened 2021-11-15
Avro schema validation order preservation

sreekesh93:


erosb (changes requested):
Hello,
thanks for raising the PR. Due to the severe performance tradeoff of `LinkedHashMap` over `HashMap` property ordering won't be acceptable for many users.

Therefore the use of `LinkedHashMap` can be only an optional thing, and this option should be turned off by default.

sreekesh93:
> Hello, thanks for raising the PR. Due to the severe performance tradeoff of `LinkedHashMap` over `HashMap` property ordering won't be acceptable for many users.
> 
> Therefore the use of `LinkedHashMap` can be only an optional thing, and this option should be turned off by default.

Will raise PR with mentioned changes, Thanks @erosb

erosb:
@sreekesh93 let me give you some ideas/clues on the implementation:

 * please create an additional flag in `SchemaLoaderBuilder` that let's the user configure if they want ordered object schemas or not
 *  this flag should be included in (and copied to, in the `SchemaLoader` constructor) `LoaderConfig` as well
 * let's create an `ObjectSchema` subclass, if possible, like `OrderedObjectSchema`  
 * let's extract all `HashMap` instantiations in `ObjectSchema` into separate package-private methods, and override these in `OrderedObjectSchema` so that they return `LinkedHashMap` instances
 * then the `ObjectSchemaExtractor` would instantiate either an `ObjectSchema` or an `OrderedObjectSchema` depending on the state of the config flag in `LoaderConfig`.

Briefly that would be it.

erosb:
no response - closing the PR.

PR #465, opened 2022-08-18
schema builder fluency

JoshMcCullough:
Replacement for: https://github.com/everit-org/json-schema/pull/390

---

**Breaking Changes**

`Schema.Builder` needs to return the subclassed builder instance rather than `Builder<S>` in order to allow for fluency while building schemas. See related issue.

Note: I also renamed `ConstSchema.ConstSchemaBuilder` to just `Builder` to match the other builders for consistency. I can pull this change back out if you don't want it. Since this PR is already a breaking change, I figured it made sense.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -167,6 +167,12 @@
                                 <sourceCompatible>true</sourceCompatible>
                             </overrideCompatibilityChangeParameter>
                             <overrideCompatibilityChangeParameter>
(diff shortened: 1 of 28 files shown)


JoshMcCullough:
@erosb I re-created this PR from a different branch. The original PR is linked for posterity. I'd still like to get this in the official repo!

PR #466, opened 2022-08-18
definition schemas

JoshMcCullough:
Adds the ability to configur definition schemas.


core/src/main/java/org/everit/json/schema/ToStringVisitor.java
```diff
@@ -227,6 +227,13 @@ private void describePropertyDependencies(Map<String, Set<String>> propertyDepen
         visit(propertyNameSchema);
     }
 
+    @Override void visitDefinitionSchemas(Map<String, Schema> definitionSchemas) {
+        if (!definitionSchemas.isEmpty()) {
+            writer.key("definitions");
```
erosb, line 232: Maybe better to use `$defs` since that is the currently recommended keyword [according to the spec](https://json-schema.org/draft/2020-12/json-schema-core.html#name-schema-re-use-with-defs)?

core/src/main/java/org/everit/json/schema/Visitor.java
```diff
@@ -177,6 +181,12 @@ void visitPatternProperties(Map<Regexp, Schema> patternProperties) {
         }
     }
 
+    void visitDefinitionSchemas(Map<String, Schema> definitionSchemas) {
```
erosb, line 184: Are you sure this makes sense? It may trigger unnecessary validation. For example if the schema is 
```json
{
   "type": "object",
   "properties": {
      "A": { "type": "string" }
   },
   "required": ["A"']
   "$defs": {
      "A": { "type": "integer" }
   }
}
```
Then I suspect no instance would pass the validation (the "A" prop would be validated against both `{"type":"string"}` and `{ "type": "integer" }`).

erosb (commented):
Thank you for raising the PR. Added a couple of notes. 
One more question: do you think it is possible to change the `SchemaLoader` so that the subschemas appearing under `"$defs"` are added to `Schema#definitionSchemas` ? I know it won't be a trivial code change, but maybe it is feasible?

Issue #395, opened 2021-01-07
Use LinkedHashMap to preserve order of schema elements


JoshMcCullough:
Currently a `HashMap` is used, so the order of e.g. properties is not preserved. A simple change to use `LinkedHashMap` instead would preserve the schema designer's insertion order.

Example:

```java
ObjectSchema.builder()
    .addProperty("one", ...)
    .addProperty("two", ...)
    .addProperty("three", ...);
```

Results in something like ...

```js
{
  "properties": [
        "two": ...,
        "three": ...,
        "one": ...,
  ]
}
```

... where the properties are in an indeterminate order (based on the hash of the key). Preferably, the order would be preserved:

```js
{
  properties: [
        one: ...,
        two: ...,
        three: ...,
  ]
}
```

[Some would argue](https://github.com/stleary/JSON-java/pull/106) "no way man, the JSON spec specifically says that property order doesn't matter" (yeah, I lost that fight and here I am again for round 2), but I'd argue that the order being effectively random is far worse than the order being preserved.

erosb:
Hello Josh,

I'm not strongly against the ordering, but I find using `LinkedHashMap` risky (there are some users who use extremely large generated schemas, and using `LinkedHashMap` could cause a significant performance loss for them). So if your usecase is only for `"properties"`, then I suggest adding an `ObjectSchemaBuilder#preserveOrder()` method that would change the internal representation from `HashMap` to `LinkedHashMap`. What do you think about that? Would it be sufficient for your usecase? (By the way I can't promise I will implement it in the near future).

JoshMcCullough:
`LinkedHashMap` generally performs better than `HashMap`, the only downside is the small amount of additional memory it would use to store the "links". I'd suggest it isn't necessary to "choose one or the other", but a straightforward performance test would answer that question.

hiloboy0119:
Hey Josh, not sure if this will help you but this is how I solved this problem:

```java
import java.util.Collections;
import java.util.LinkedHashMap;
import java.util.Map;
import org.everit.json.schema.ObjectSchema;
import org.everit.json.schema.Schema;

public class OrderedObjectSchema extends ObjectSchema {

  public static class Builder extends ObjectSchema.Builder {
    private final Map<String, Schema> orderedPropertySchemas = new LinkedHashMap<>();

    public Builder addOrderedPropertySchema(String propName, Schema schema) {
      super.addPropertySchema(propName, schema);
      orderedPropertySchemas.put(propName, schema);
[...]

PR #467, opened 2022-08-18
use linked hash maps where it matters

JoshMcCullough:
Fixes https://github.com/everit-org/json-schema/issues/395.


erosb (changes requested):
Hello, using `LinkedHashMap` should be an optional feature which is disabled by default, due to possible performance degradation in some usecases (`LinkedHashMap` is slower than `HashMap`). So please make ordering toggle-able in `SchemaLoaderBuilder`, and instantiate `HashMap` / `LinkedHashMap` depending on the feature toggle during the schema loading process. For that you can start looking at `ObjectSchemaLoader:59`.
I think it is enough for now to implement order-keeping for `"properties"`.

PR #501, opened 2023-12-15
Proposed fix reference that cause StackOverflowException on equals/hashCode

tledkov:


erosb (approved):
I'm a little bit concerned about potential BC breaks here, but I don't have better idea either, so let's go with it. Thanks!

PR #519, opened 2025-02-01
Add ability to get subschemas of CombinedSchema in insertion order

rayokota:
Add the ability to get subschemas of CombinedSchema in insertion order.

For context, the code in [Confluent Schema Registry](https://github.com/confluentinc/schema-registry) is currently using version 1.14.3 of this library.  We found that when we upgraded to 1.14.4 some of our code no longer worked because we were relying on the order of schemas returned from `CombinedSchema.getSubschemas()`.  That order changed in 1.14.4 due to https://github.com/everit-org/json-schema/pull/498.  It turns out the order was not deterministic anyway for the leading subschemas, but we hadn't yet hit this problem.  But now while the order is deterministic for any one specific JVM execution, it may be different between different executions and different JVMs because it uses the `hashCode` method to sort the subschemas.

Our code works best with the insertion order, which is deterministic (across executions and JVMs) for a given `CombinedSchema`.  Therefore a modest workaround that works for us is to add a `CombinedSchema.getOrderedSubschemas` method and for the Confluent code to rely on that method instead of `CombinedSchema.getSubschemas`.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/CombinedSchema.java
+++ b/core/src/main/java/org/everit/json/schema/CombinedSchema.java
@@ -154,6 +154,8 @@ public class CombinedSchema extends Schema {
 
     private final boolean synthetic;
 
+    private final Collection<Schema> orderedSubschemas;
+
     private final Collection<Schema> subschemas;
 
     private final ValidationCriterion criterion;
@@ -168,6 +170,7 @@ public class CombinedSchema extends Schema {
         super(builder);
         this.synthetic = builder.synthetic;
         this.criterion = requireNonNull(builder.criterion, "criterion cannot be null");
+        this.orderedSubschemas = builder.subschemas;
         this.subschemas = sortByCombinedFirst(requireNonNull(builder.subschemas, "subschemas cannot be null"));
     }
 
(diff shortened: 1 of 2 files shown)


erosb:
Hello @rayokota  I thought you already migrated to erosb/json-sKema, didn't you?

rayokota:
Hi @erosb , we still use this library for the older drafts, and also for the object model it provides.

erosb:
Hello @rayokota does this issue have  an originator issue in the schema registry GH repo?

rayokota:
Thanks @erosb , I've filed https://github.com/confluentinc/schema-registry/issues/3536 to track.  I'm happy to answer any further questions you might have.

rayokota:
Hi @erosb , do you have any more feedback/questions on this PR?

erosb:
Hello @rayokota I'm thinking about if it is possible to fix this issue without introducing an other public method on CombinedSchema. I pushed a [commit](https://github.com/everit-org/json-schema/commit/13a9ea13233032a8cc405bcad622203d5d1f8b82) into a separate branch, trying to implement a comparator which remembers the original ordering of subschemas, but so far it doesn't work (the `CombinedSchemaTest#subschemasSort()` fails on this change).

rayokota:
Thanks for looking into this @erosb !  I really appreciate it.  

Anyway, I have an alternative API that does not introduce another public method, but simply restores the insertion order (which used to be the original behavior): https://github.com/everit-org/json-schema/pull/522

Issue #520, opened 2025-02-01
Schema with multiple anyOf subschemas throws `org.json.JSONException: Duplicate key "anyOf"`


rayokota:
When the schema below is loaded and then `toString()` is called, then `org.json.JSONException: Duplicate key "anyOf"` results.
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "updatedAt": {
      "type": [
        "string",
        "null"
      ],
      "anyOf": [
        {
          "format": "date-time"
        },
        {
          "format": "date"
        }
      ]
    }
  }
}
```

The reason is that the "type" element is also converted into an "anyOf", which is a duplicate key to the "anyOf" with the format elements.

I will submit a PR with a possible fix.

erosb:
Thank you for the PR @rayokota , looks good. I also don't have of any better way to fix it, so let's go ahead with merging.

rayokota:
Thanks for the quick turnaround @erosb !

PR #521, opened 2025-02-01
Handle multiple anyOf subschemas in ToStringVisitor

rayokota:
When converting a Schema to a string, if a synthetic allOf schema has multiple subschemas of the same criterion, then emit the allOf to prevent `org.json.JSONException: Duplicate key "anyOf"`

Fixes https://github.com/everit-org/json-schema/issues/520


rayokota:
@erosb , please feel free to modify this solution or ditch it for something better :)

PR #522, opened 2025-02-05
Maintain insertion order for subschemas of CombinedSchema

rayokota:
Maintain insertion order for subschemas of CombinedSchema.

This is an alternative to https://github.com/everit-org/json-schema/pull/519 that maintains the subschemas in insertion order.

The subschemas were originally maintained in insertion order, but it was changed due to https://github.com/everit-org/json-schema/pull/405, which ensured that subschemas that were of type CombinedSchema were visited first during validation, and by https://github.com/everit-org/json-schema/pull/498, which tries to maintain a stable order for "equivalent" instances.

This PR simply restores the insertion order, but maintains an internally sorted collection for the two purposes listed above, namely visiting subschemas during validation and equivalence checks during equals/hashCode.

### DIFF
--- a/core/src/main/java/org/everit/json/schema/CombinedSchema.java
+++ b/core/src/main/java/org/everit/json/schema/CombinedSchema.java
@@ -156,6 +156,8 @@ public class CombinedSchema extends Schema {
 
     private final Collection<Schema> subschemas;
 
+    private final Collection<Schema> sortedSubschemas;
+
     private final ValidationCriterion criterion;
 
     /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/everit/json/schema/CombinedSchema.java
```diff
@@ -191,10 +194,22 @@ public ValidationCriterion getCriterion() {
...
+     */
     public Collection<Schema> getSubschemas() {
         return subschemas;
     }
 
+    /*
+     * Internal method that returns the subschemas in the order they should be visited
+     * by the ValidatingVisitor and the equals and hashCode methods.
+     */
+    protected Collection<Schema> visitSubschemas() {
```
erosb, line 209: nit: I think package-private visibility would be better here, so that potential subclasses outside the package don't see it. It would also mean, we wouldn't need to maintain backward compatibility in the future - just in case we change our mind.

Also some little more descriptive would be welcome. How about `subschemasWithCombinedFirst()` ?

rayokota: Thanks, incorporated your suggestions.

erosb (approved):
Then let's go ahead with this one :)

rayokota:
Thanks again @erosb ! I really appreciate all your efforts!  :)

erosb:
Welcome :) version 1.14.5 is already available on maven central with your fix.

